PR #2, opened 2023-05-30
DistributedArray with functions

rohanbabbar04:
- `DistributedArray` class
 - Offers global array to be replicated at all ranks, global array divided into parts.
 - `to_dist` method which converts the global array to a DistributedArray
 - Added add, subtract, and multiplication in a distributed fashion.
 - Tests and examples

### DIFF
--- a/pylops_mpi/DistributedArray.py
+++ b/pylops_mpi/DistributedArray.py
@@ -25,7 +25,7 @@ def local_split(global_shape: Tuple, base_comm: MPI.Comm, partition: str):
     """
     if partition == "B":
         local_shape = global_shape
-    # Scatter the array
+    # Split the array
     else:
         local_shape = [((global_shape[0] // base_comm.Get_size()) + 1)
                        if base_comm.Get_rank() < (
@@ -48,7 +48,7 @@ class DistributedArray:
         MPI Communicator over which array is distributed.
         Defaults to ``MPI.COMM_WORLD``.
     partition : :obj:`str`, optional
-        Broadcast or Scatter the array. Defaults to ``S``.
+        Broadcast or Split the array. Defaults to ``S``.
     dtype : :obj:`str`, optional
         Type of elements in input array. Defaults to ``numpy.float64``.
     """


setup.py
```diff
@@ -0,0 +1,39 @@
...
+setup(
+    name='pylops_mpi',
+    description=descr,
+    long_description=open(src('README.md')).read(),
+    long_description_content_type='text/markdown',
+    keywords=['algebra',
+              'inverse problems',
+              'large-scale optimization'],
+    classifiers=[
+        'Development Status :: 5 - Production/Stable',
```
mrava87, line 22: Change this into `Development Status :: 1 - Planning` for now

pylops_mpi/DistributedArray.py
```diff
@@ -0,0 +1,245 @@
+import numpy as np
+from typing import Optional, Union, Tuple
+from numbers import Integral
+from mpi4py import MPI
+
+from pylops.utils import DTypeLike, NDArray
+
+
+def local_split(global_shape: Tuple, base_comm: MPI.Comm, partition: str):
```
mrava87, line 9: Please add docstring

mrava87 (commented):
Good job. I gave you some specific comments and will give you a few more general ones in a separate post.

pylops_mpi/DistributedArray.py
```diff
@@ -0,0 +1,245 @@
...
+
+    def get_local_arrays(self):
+        """Gather all the local arrays
+        Returns
+        -------
+        final_array : :obj:`np.ndarray`
+            Global Array gathered at all ranks
+        """
+        final_array = self.base_comm.allgather(self.local_array)
+        return np.asarray(final_array)
```
mrava87, line 140: Not sure this is needed. I currently get the following warning
```
VisibleDeprecationWarning: Creating an ndarray from ragged nested sequences (which is a list-or-tuple of lists-or-tuples-or ndarrays with different lengths or shapes) is deprecated. If you meant to do this, you must specify 'dtype=object' when creating the ndarray.
```

and I also see that in `asarray` you use `np.concatenate` on the output of `get_local_arrays`... maybe get local arrays could already use that?

tests/__init__.py
hongyx11, line 1: A stupid question: Why `tests` need `__init__.py`?

tests/test_distributedarray.py
```diff
@@ -0,0 +1,101 @@
+"""Test the DistributedArray class
+    Designed to run with n processes
+    $ mpiexec -n 10 pytest -m mpi test_distributedarray.py
```
hongyx11, line 3:
```
  /home/hongy0a/pylops-mpi/tests/test_distributedarray.py:27: PytestUnknownMarkWarning: Unknown pytest.mark.mpi - is this a typo?  You can register custom marks to avoid this warning - for details, see https://docs.pytest.org/en/stable/how-to/mark.html
    @pytest.mark.mpi(min_size=2) 
```
currently I got this warning when launching this test file.

setup.py
rohanbabbar04: Done...

pylops_mpi/DistributedArray.py
rohanbabbar04: Added the docstrings

pylops_mpi/DistributedArray.py
rohanbabbar04: Merged the code of both into `asarray`, I thinks its a better name and will be used to call the global view of the entire array.

tests/__init__.py
rohanbabbar04: True, no need for `__init__.py`, deleted the file :slightly_smiling_face:

tests/test_distributedarray.py
rohanbabbar04: I think you do not have pytest-mpi installed
just do a `pip install pytest-mpi` or just install all the requirements using `pip install -r requirements-dev.txt`
Then either use `make tests` or use `mpiexec -n 10 pytest test_distributedarray.py --with-mpi`
Then it will work accordingly...

pylops_mpi/DistributedArray.py
mrava87: Good idea :)

tests/test_distributedarray.py
mrava87: I had done the same mistake initially (using an old env). I confirm this works :)

mrava87:
@rohanbabbar04 I think we can merge this if you agree, so we dont make the PR too big, and can always go back later to improve some of the pieces you have added to the package :)

PR #7, opened 2023-06-02
DistributedArray along an axis and plotting functions for better examples

rohanbabbar04:
Enhancement for Distribution of array along an axis.
- [x] Code in DistributedArray class
- [x] Updated the Tests
- [x] Added Plotting for DistributedArray (aligns with pylops)
- [x]  Update the Examples


rohanbabbar04:
- Added the code for `DistributedArray` split along an axis. Added `axis` as a member
- Updated the Tests accordingly
- For the example, I have switched from displaying arrays to plotting using `matplotlib`(just like we do in pylops)
- Added a function `plot_distributed_array` to achieve it.

Issue #16, opened 2023-06-10
README


mrava87:
Start writing a REAMDE file along the same lines as PyLops one. For now it can be rather short and should simply include the main details about:

- [x] Project idea
- [x] Installation
- [x] A simple motivating example (for now just about the DistributedArray.. will change later)
- [x] How to run a PyLops-MPI code (given that this is not anymore just python file.py, it is good to give some guidelines)
- [x] Running tests
- [x] Documentation
- [x] Contributors

PR #22, opened 2023-06-12
Updated Readme

rohanbabbar04:
For #16 
- Updated the README.md
- As in the [Discussion #1](https://github.com/PyLops/pylops-mpi/discussions/1), Changed the shape to (10, 5) in example, also changed the shape in tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # pylops-mpi
 [![pages-build-deployment](https://github.com/PyLops/pylops-mpi/actions/workflows/pages/pages-build-deployment/badge.svg)](https://github.com/PyLops/pylops-mpi/actions/workflows/pages/pages-build-deployment)
 [![PyLops-MPI](https://github.com/PyLops/pylops-mpi/actions/workflows/build.yml/badge.svg)](https://github.com/PyLops/pylops-mpi/actions/workflows/build.yml)
-![OS-support](https://img.shields.io/badge/OS-linux,win,osx-850A8B.svg)
+![OS-support](https://img.shields.io/badge/OS-linux,osx-850A8B.svg)
 [![Slack Status](https://img.shields.io/badge/chat-slack-green.svg)](https://pylops.slack.com)
 
 ## PyLops MPI
@@ -22,12 +22,12 @@ and running the following command:
    ```
  3. **Install pylops-mpi**: Once MPI is installed and verified, you can proceed to install `pylops-mpi`. 
    
-   You can install with `make` and `pip`
+   You can install with `make` and `pip`:
       ```
       make install
       ```
    
-   You can install with `make` and `conda`
+   You can install with `make` and `conda`:
       ```
       make install_conda
       ```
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1,2 +1,93 @@
 # pylops-mpi
-MPI-powered PyLops with MPI4Py  Topics Resources - GSOC2023 Project
+[![pages-build-deployment](https://github.com/PyLops/pylops-mpi/actions/workflows/pages/pages-build-deployment/badge.svg)](https://github.com/PyLops/pylops-mpi/actions/workflows/pages/pages-build-deployment)
+[![PyLops-MPI](https://github.com/PyLops/pylops-mpi/actions/workflows/build.yml/badge.svg)](https://github.com/PyLops/pylops-mpi/actions/workflows/build.yml)
+![OS-support](https://img.shields.io/badge/OS-linux,win,osx-850A8B.svg)
```
mrava87, line 4: Small comment, unlikely this will work on windows... maybe for now we should just have linux as I suspect this is where you develop (and where tests are run)?

README.md
```diff
@@ -1,2 +1,93 @@
...
+    mpiexec --version
+   ```
+ 3. **Install pylops-mpi**: Once MPI is installed and verified, you can proceed to install `pylops-mpi`. 
+   
+   You can install with `make` and `pip`
+      ```
+      make install
+      ```
+   
+   You can install with `make` and `conda`
```
mrava87, line 30: add `:` as right now the command is not separated and may lead to some confusion (same for the pip one above)

README.md
```diff
@@ -1,2 +1,93 @@
...
+The DistributedArray also provides fundamental mathematical operations, like element-wise addition, subtraction, and multiplication, 
+as well as dot product and the [`np.linalg.norm`](https://numpy.org/doc/stable/reference/generated/numpy.linalg.norm.html) function in a distributed fashion, 
+thus utilizing the efficiency of the MPI protocol. This enables efficient computation and processing of large-scale distributed arrays.
+
+## Documentation 
+The official documentation of Pylops-MPI is available [here](https://pylops.github.io/pylops-mpi/).
+Visit the official docs to learn more about pylops-mpi.
+
+## Contributors
+* Matteo Ravasi, mrava87
```
mrava87, line 92: You should be first, both because of alphabetic order, second because of size of contribution ;)

mrava87 (commented):
Good stuff! 

The test/example changes look good (nice this did not lead to any unexpected bug ;) ).
The README looks great and informative.

I would just one more small section on how to run tests, other than that I am happy to approve and let you merge once ready.

README.md
rohanbabbar04: As the tests are working on both mac and ubuntu, it does work on both.
For windows I think it should work but will have to check.
Changed to `linux, osx` as of now...

README.md
rohanbabbar04: Done...

README.md
rohanbabbar04: Done :+1:

rohanbabbar04:
Added the running tests section to the readme

mrava87 (commented):
Great, good job :)

PR #24, opened 2023-06-17
Display gallery examples using mpi

rohanbabbar04:
Like we had discussed..
- Added the `note` to `examples/README.rst`
- Added mpi_examples.sh which runs all the files in the folder, one can run using `sh mpi_examples.sh <num_processes>`
- Also added `run_examples` in the `Makefile`.
- Also changed the `NUM_PROCESSES` in `Makefile` to 3 for quicker results..

### DIFF
--- a/examples/README.rst
+++ b/examples/README.rst
@@ -7,3 +7,9 @@ Here is a collection of examples that demonstrate the usage of PyLops-MPI operat
     All examples are currently run on a single rank when building the documentation.
     However, we test their correctness off-line with multiple ranks and we encourage users to also
     run them with multiple ranks when learning about pylops-mpi.
+
+    Use the following command to run the gallery examples using MPI.
+
+    .. code-block:: shell-session
+
+       $ make run_examples
--- a/examples/README.rst
+++ b/examples/README.rst
@@ -1,4 +1,9 @@
 ===============
 Gallery
 ===============
-Examples using pylops_mpi
+Here is a collection of examples that demonstrate the usage of PyLops-MPI operators and utilities.
+
+.. note::
+    All examples are currently run on a single rank when building the documentation.
+    However, we test their correctness off-line with multiple ranks and we encourage users to also
+    run them with multiple ranks when learning about pylops-mpi.
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 PIP := $(shell command -v pip3 2> /dev/null || command which pip 2> /dev/null)
 PYTHON := $(shell command -v python3 2> /dev/null || command which python 2> /dev/null)
-NUM_PROCESSES = 5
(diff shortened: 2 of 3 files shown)


mrava87:
@rohanbabbar04 What is this?

mrava87:
>Right I see. So what is it doing right now? And how\why are the plot worse if not run with `mpiexec`?

rohanbabbar04:
> > Right I see. So what is it doing right now? And how\why are the plot worse if not run with `mpiexec`?

Right now it is taking into consideration only one rank, so we can see only for rank 0, but would be good if the py files run with more than 1

mrava87:
Oh yes, that would be ideal as otherwise it isn't really running in ad distributed fashion. What problem are you facing?

rohanbabbar04:
The problem is that I can't really find a way to run the python files(example files that sphinx runs) using mpiexec
If you find any resource, plz share it

mrava87:
See https://github.com/sphinx-gallery/sphinx-gallery/issues/1149

For now I suggest we just run the examples with one rank, but also run them manually or via a sh script to make sure they run with n ranks and produce meaningful results (to avoid what happened in the poststack case). We can later investigate if we can find an alternative route :)

rohanbabbar04:
> See [sphinx-gallery/sphinx-gallery#1149](https://github.com/sphinx-gallery/sphinx-gallery/issues/1149)
> 
> For now I suggest we just run the examples with one rank, but also run them manually or via a sh script to make sure they run with n ranks and produce meaningful results (to avoid what happened in the poststack case). We can later investigate if we can find an alternative route :)

Thanks for this, I'll get onto it...

mrava87:
Great!

For the moment, we can perhaps also add a comment here https://github.com/PyLops/pylops-mpi/blob/main/examples/README.rst by saying `Note that all examples are currently run on a single rank when building the documentation. However, we test their correctness off-line with multiple ranks and we encourage users to also run them with multiple ranks when learning about pylops-mpi.`

examples/README.rst
```diff
@@ -1,4 +1,9 @@
 ===============
 Gallery
 ===============
-Examples using pylops_mpi
+Here is a collection of examples that demonstrate the usage of PyLops-MPI operators and utilities.
+
+.. note::
+    All examples are currently run on a single rank when building the documentation.
+    However, we test their correctness off-line with multiple ranks and we encourage users to also
+    run them with multiple ranks when learning about pylops-mpi.
```
mrava87, line 9: since you made a new command in the Makefile, makes sense to either add a comment about it here or in the main README under the doc section

rohanbabbar04: I have added it over here, I think it would be better as it is for gallery examples....

mrava87 (approved):
Very good

PR #25, opened 2023-06-26
Added MPILinearOperator and MPIBlockdiag

rohanbabbar04:
- Added `local_shapes` as a parameter in the DistributedArray to force ranks for a particular shape before assigning the array.
- Added `MPILinearOperator` similar to `pylops-mpi-old`, added the functionality of Distribution using DistributedArray along with all the important ops such as `_AdjointLinearOperator`, `_SumLinearOperator`, etc..
- Added `MatrixMult`, _matvec and _rmatvec using DistributedArray as it will be an important op, also will need handling as it uses self.A.
- Added  `MPIBlockDiag`, currently each operator is assigned to a rank and computation happens at each rank..
- Added tests for the above

### DIFF
--- a/pylops_mpi/BlockDiag.py
+++ b/pylops_mpi/BlockDiag.py
@@ -14,6 +14,7 @@ class MPIBlockDiag(MPILinearOperator):
 
     def __init__(self, ops, base_comm: MPI.Comm = MPI.COMM_WORLD, dtype: Optional[DTypeLike] = None):
         self.ops = self._assign_ops(ops, base_comm)
+        print(self.ops)
         self.kind = "mix"
         mops = np.zeros(len(self.ops), dtype=np.int64)
         nops = np.zeros(len(self.ops), dtype=np.int64)
@@ -37,20 +38,22 @@ class MPIBlockDiag(MPILinearOperator):
         x = np.insert(np.cumsum(base_comm.allgather(local_shapes)), 0, 0)
         return ops[x[base_comm.Get_rank()]:x[base_comm.Get_rank() + 1]]
 
-    def _matvec(self, x: Union[NDArray, DistributedArray]):
+    def _matvec(self, x):
         y = DistributedArray(global_shape=self.shape[0], local_shapes=self.nops)
         if not isinstance(x, DistributedArray):
             x = DistributedArray.to_dist(x=x, local_shapes=self.mops)
+        print(x)
         y1 = []
         for iop, oper in enumerate(self.ops):
             if not isinstance(oper, MPILinearOperator):
                 oper = asmpilinearoperator(Op=oper)
-            y1.append(oper.matvec(x.local_array[self.mmops[iop]: self.mmops[iop+1]]))
+            y1.append(oper.matvec(x.local_array[self.mmops[iop]:
+                                                self.mmops[iop + 1]]))
         if y1:
             y[:] = np.concatenate(y1)
         return y
 
-    def _rmatvec(self, x: Union[NDArray, DistributedArray]):
+    def _rmatvec(self, x):
         y = DistributedArray(global_shape=self.shape[1], local_shapes=self.mops)
         if not isinstance(x, DistributedArray):
             x = DistributedArray.to_dist(x=x, local_shapes=self.nops)
@@ -62,4 +65,3 @@ class MPIBlockDiag(MPILinearOperator):
         if y1:
             y[:] = np.concatenate(y1)
         return y
-
--- a/pylops_mpi/BlockDiag.py
+++ b/pylops_mpi/BlockDiag.py
@@ -1,8 +1,8 @@
 import numpy as np
 from mpi4py import MPI
-from typing import Optional, Union
+from typing import Optional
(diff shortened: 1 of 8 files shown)


rohanbabbar04:
I had also looked at `pylops-old-mpi` and found that there was an implementation of `BlockDiag`, where we give an option to operator to decide whether to go `all` or `master`. I looked into that , below is a code for that, My only problem is that for some cases especially non-square, some issues are popping. Should I implement the below one or keep it as it is(assinging different ops and data to different ranks)...
```
from pylops_mpi import MPILinearOperator
import numpy as np
import scipy as sp
from mpi4py import MPI
from typing import Optional
from pylops.utils import DTypeLike

sp_version = sp.__version__.split(".")
if int(sp_version[0]) <= 1 and int(sp_version[1]) < 8:
    from scipy.sparse.linalg.interface import _get_dtype
else:
    from scipy.sparse.linalg._interface import _get_dtype


class MPIBlockDiag(MPILinearOperator):

    def __init__(self, ops, base_comm: MPI.Comm = MPI.COMM_WORLD, dtype: Optional[DTypeLike] = None):
        self.ops = ops
        self.kind = "mix"
        mops = np.zeros(len(self.ops), dtype=np.int64)
        nops = np.zeros(len(self.ops), dtype=np.int64)
        self.mmops = []
        self.nnops = []
        for iop, oper in enumerate(self.ops):
            if oper.kind == "master":
                # Handle mmops and nnops for master
                if base_comm.rank == 0:
                    nops[iop] = oper.shape[0]
                    mops[iop] = oper.shape[1]
                self.mmops.append(oper.shape[1])
                self.nnops.append(oper.shape[0])
            else:
                # For operators that are distributed among ranks
                nops[iop] = oper.shape[0]
                mops[iop] = oper.shape[1]
                self.nnops.append(np.sum(base_comm.allgather(oper.shape[0])))
                self.mmops.append(np.sum(base_comm.allgather(oper.shape[1])))
        self.mops = mops.sum()
        self.nops = nops.sum()
        self.nnops = np.insert(np.cumsum(self.nnops), 0, 0)
        self.shape = (nops.sum(), mops.sum())
        if dtype:
            self.dtype = _get_dtype(dtype)
        else:
            self.dtype = np.dtype(dtype)
        super().__init__(shape=self.shape, dtype=self.dtype, base_comm=base_comm, kind=self.kind)

    def _matvec(self, x):
        y1 = []
        for iop, oper in enumerate(self.ops):
            y = oper.matvec(x[self.mmops[iop]: self.mmops[iop + 1]])
            if y is not None:
                y1.append(y)
        return np.hstack(y1)

    def _rmatvec(self, x):
        y1 = []
        for iop, oper in enumerate(self.ops):
            y = oper.rmatvec(x[self.nnops[iop]: self.nnops[iop + 1]])
            if y is not None:
                y1.append(y)
        return np.hstack(y1)

```

mrava87:
Hi @rohanbabbar04,
Thanks!

I am back from holiday tomorrow and will try to look into this before the end of the week.

I may however suggest you make a couple go separate PRs as this is very big and would be difficult to review/comment/amend without getting too many things to overlap.

Let’s focus first on MPIBlockDiag as it was it the roadmap. For now assigning each operator to a rank is perfect, this is how I would see this working. If one wants to assign a group of operators could just pass BlockDiag to MPIBlockDiag :)

I don’t see anymore the question you made about pylops-mpi-old, but the reply is no. The master/all need arose from the specific problem I started to look into (which is the family 3 of problems in our proposal). We agreed this will be tackled last if time alllows so let’s forget about it for now.

At the moment I’m also not sure why MPILinearOperator is needed. Again we didn’t discuss about working on this so I would suggest parking it for the moment and focus on the operators we have in the proposal timeline. I will take a look at it anyways, and give you a more detailed feedback…

mrava87:
Also, I currently do not see any example.

Please try in the meantime to focus on making some basic example and a first tutorial of MPIBlockDiag as we discussed in our kick off meeting. Use this https://pylops.readthedocs.io/en/stable/tutorials/poststack.html#sphx-glr-tutorials-poststack-py as the problem we would like to tackle (for now only the forward and adjoint as we miss the solver to run the inversion). In the slides I made there is also a link to a notebook where I made a simple 3d example just replicating along the third direction, you can use that as starting point if you prefer :)

rohanbabbar04:
Fine 
I'll make smaller PR's
Let's first get `MPIBlockDiag` in then look into others..

mrava87:
A few additional general comment:

- try to use the same folder structure as pylops. So BlockDiag and MatrixMult should go into a basicoperators folders
- please comment as you go, right now MPIBlockDiag has no docstring nor comments… this makes it harder for us to review

rohanbabbar04:
Just wanted a review/suggestion
Will make smaller PRs for the above with comments...

mrava87:
> Just wanted a review/suggestion
> Will make smaller PRs for the above with comments...

Alright, then I think for now you have a few general pointers, more detailed comments coming soon (here or on the new PRs :) )

PR #26, opened 2023-06-27
Add MPIBlockDiag and MPILinearOperator

rohanbabbar04:
- Added MPIBlockDiag with _matvec and _rmatvec which now uses `DistributedArray`
- Added MPILinearOperator class 
- Added basic classes(_ScaledLinearOperator, _AdjointLinearOperator, etc) and made them input and output `DistributedArray`.
- Added tests to verify above classes.
- Upgrade scipy to `1.8` and python to `3.8`
- Updated docs for sphinx

### DIFF
--- a/pylops_mpi/LinearOperator.py
+++ b/pylops_mpi/LinearOperator.py
@@ -1,0 +1,332 @@
+import numpy as np
+from mpi4py import MPI
+from typing import Callable
+
+from scipy.sparse._sputils import isintlike
+from scipy.sparse.linalg._interface import _get_dtype
+
+from pylops import LinearOperator
+from pylops.utils import DTypeLike, ShapeLike
+
+from pylops_mpi import DistributedArray
+
+
+class MPILinearOperator(LinearOperator):
+    """Common interface for performing matrix-vector products in distributed fashion.
+
+    This class extends the :class:`pylops.LinearOperator`.
+    This class provides methods to perform matrix-vector product and adjoint matrix-vector
+    products using MPI.
+
+    .. note:: End users of pylops-mpi should not use this class directly but simply
+      use operators that are already implemented. This class is meant for
+      developers only, it has to be used as the parent class of any new operator
+      developed within pylops-mpi.
+
+    Parameters
+    ----------
+    shape : :obj:`tuple(int, int)`
+        Shape of the MPI Linear Operator.
+    dtype : :obj:`str`
+        Type of elements in input array.
+    Op : :obj:`pylops.LinearOperator`, optional
+        Linear Operator. Defaults to ``None``.
+    base_comm : :obj:`mpi4py.MPI.Comm`, optional
+        MPI Base Communicator. Defaults to ``mpi4py.MPI.COMM_WORLD``.
+
+    """
+
+    def __init__(self, shape: ShapeLike, dtype: DTypeLike, Op: LinearOperator = None,
+                 base_comm: MPI.Comm = MPI.COMM_WORLD):
+        self.Op = Op
+        # For MPI
+        self.base_comm = base_comm
+        self.size = self.base_comm.Get_size()
+        self.rank = self.base_comm.Get_rank()
+        super().__init__(Op=Op, dtype=dtype, shape=shape)
+
+    def matvec(self, x: DistributedArray) -> DistributedArray:
+        if self.Op:
+            y = DistributedArray(global_shape=self.base_comm.allreduce(self.shape[1]), dtype=x.dtype)
+            if isinstance(x, DistributedArray):
+                x = x.local_array
+            y[:] = self.Op._matvec(x)
+        else:
+            y = self._matvec(x)
+        return y
+
+    def rmatvec(self, x: DistributedArray) -> DistributedArray:
+        if self.Op:
+            y = DistributedArray(global_shape=self.base_comm.allreduce(self.shape[1]), dtype=x.dtype)
+            if isinstance(x, DistributedArray):
+                x = x.local_array
+            y[:] = self.Op._rmatvec(x)
+        else:
+            y = self._rmatvec(x)
+        return y
+
+    def dot(self, x: DistributedArray):
+        if isinstance(x, MPILinearOperator):
+            Op = _ProductLinearOperator(self, x)
+            Op.clinear = (getattr(self, 'clinear', True)
+                          and getattr(x, 'clinear', True))
+            return Op
+        elif np.isscalar(x):
+            return _ScaledLinearOperator(self, x)
+        else:
+            if x is None or x.ndim == 1:
+                return self.matvec(x)
+            elif x.ndim == 2:
+                return self.matmat(x)
+            else:
+                raise ValueError('expected 1-d or 2-d array or matrix, got %r'
+                                 % x)
+
+    def adjoint(self):
+        return self._adjoint()
+
+    H = property(adjoint)
+
+    def transpose(self):
+        return self._transpose()
+
+    T = property(transpose)
+
+    def __mul__(self, x):
+        return self.dot(x)
+
+    def __rmul__(self, x):
+        if np.isscalar(x):
+            return _ScaledLinearOperator(self, x)
+        else:
+            return NotImplemented
+
+    def __matmul__(self, x):
+        if np.isscalar(x):
+            raise ValueError("Scalar not allowed, use * instead")
+        return self.__mul__(x)
+
+    def __rmatmul__(self, x):
+        if np.isscalar(x):
+            raise ValueError("Scalar not allowed, use * instead")
+        return self.__rmul__(x)
+
+    def __pow__(self, p):
+        return _PowerLinearOperator(self, p)
+
+    def __add__(self, x):
+        return _SumLinearOperator(self, x)
+
+    def __neg__(self):
+        return _ScaledLinearOperator(self, -1)
+
+    def __sub__(self, x):
+        return self.__add__(-x)
+
+    def _adjoint(self):
+        return _AdjointLinearOperator(self)
+
+    def _transpose(self):
+        return _TransposedLinearOperator(self)
+
+    def conj(self):
+        return _ConjLinearOperator(self)
+
+
+class _AdjointLinearOperator(MPILinearOperator):
+    """Adjoint of MPI Linear Operator"""
+
+    def __init__(self, A: MPILinearOperator):
+        self.A = A
+        self.args = (A,)
+        super().__init__(shape=(A.shape[1], A.shape[0]), dtype=A.dtype,
+                         base_comm=MPI.COMM_WORLD)
+
+    def _matvec(self, x: DistributedArray):
+        return self.A.rmatvec(x)
+
+    def _rmatvec(self, x: DistributedArray):
+        return self.A.matvec(x)
+
+
+class _TransposedLinearOperator(MPILinearOperator):
+    """Transposition of MPI Linear Operator"""
+
+    def __init__(self, A: MPILinearOperator):
+        self.A = A
+        self.args = (A,)
+        super().__init__(shape=(A.shape[1], A.shape[0]), dtype=A.dtype,
+                         base_comm=MPI.COMM_WORLD)
+
+    def _matvec(self, x: DistributedArray):
+        x[:] = np.conj(x.local_array)
+        y = self.A.rmatvec(x)
+        y[:] = np.conj(y.local_array)
+        return y
+
+    def _rmatvec(self, x: DistributedArray):
+        x[:] = np.conj(x.local_array)
+        y = self.A.matvec(x)
+        y[:] = np.conj(y.local_array)
+        return y
+
+
+class _ProductLinearOperator(MPILinearOperator):
+    """Product of MPI LinearOperators
+    """
+    def __init__(self, A: MPILinearOperator, B: MPILinearOperator):
+        if not isinstance(A, MPILinearOperator) or not isinstance(B, MPILinearOperator):
+            raise ValueError('both operands have to be a LinearOperator')
+        if A.shape[1] != B.shape[0]:
+            raise ValueError('cannot multiply %r and %r: shape mismatch' % (A, B))
+        self.args = (A, B)
+        super().__init__(shape=(A.shape[0], B.shape[1]), dtype=_get_dtype([A, B]),
+                         base_comm=MPI.COMM_WORLD)
+
+    def _matvec(self, x: DistributedArray):
+        return self.args[0].matvec(self.args[1].matvec(x))
+
+    def _rmatvec(self, x: DistributedArray):
+        return self.args[1].rmatvec(self.args[0].rmatvec(x))
+
+    def _adjoint(self):
+        A, B = self.args
+        return B.H * A.H
+
+
+class _ScaledLinearOperator(MPILinearOperator):
+    """Scaled MPI Linear Operator
+    """
+
+    def __init__(self, A: MPILinearOperator, alpha):
+        if not isinstance(A, MPILinearOperator):
+            raise ValueError('MPILinearOperator expected as A')
+        if not np.isscalar(alpha):
+            raise ValueError('scalar expected as alpha')
+        self.args = (A, alpha)
+        super().__init__(shape=A.shape, dtype=_get_dtype([A], [type(alpha)]),
+                         base_comm=MPI.COMM_WORLD)
+
+    def _matvec(self, x: DistributedArray):
+        y = self.args[0].matvec(x)
+        if y is not None:
+            y[:] *= self.args[1]
+        return y
+
+    def _rmatvec(self, x: DistributedArray):
+        y = self.args[0].rmatvec(x)
+        if y is not None:
+            y[:] *= np.conj(self.args[1])
+        return y
+
+    def _adjoint(self):
+        A, alpha = self.args
+        return A.H * np.conj(alpha)
+
+
+class _SumLinearOperator(MPILinearOperator):
+    """Sum of MPI LinearOperators
+    """
+
+    def __init__(self, A: MPILinearOperator, B: MPILinearOperator):
+        if not isinstance(A, MPILinearOperator) or not isinstance(B, MPILinearOperator):
+            raise ValueError('both operands have to be a MPILinearOperator')
+        # Make sure it works with different kinds
+        if A.shape != B.shape:
+            raise ValueError("cannot add %r and %r: shape mismatch" % (A, B))
+        self.args = (A, B)
+        super().__init__(shape=A.shape, dtype=A.dtype, base_comm=MPI.COMM_WORLD)
+
+    def _matvec(self, x: DistributedArray):
+        arr1 = self.args[0].matvec(x)
+        arr2 = self.args[1].matvec(x)
+        return arr1 + arr2
+
+    def _rmatvec(self, x):
+        arr1 = self.args[0].rmatvec(x)
+        arr2 = self.args[1].rmatvec(x)
+        return arr1 + arr2
+
+    def _adjoint(self):
+        A, B = self.args
+        return A.H + B.H
+
+
+class _PowerLinearOperator(MPILinearOperator):
+    """Power of MPI Linear Operator
+    """
+
+    def __init__(self, A: MPILinearOperator, p: int) -> None:
+        if not isinstance(A, MPILinearOperator):
+            raise ValueError("LinearOperator expected as A")
+        if A.shape[0] != A.shape[1]:
+            raise ValueError("square LinearOperator expected, got %r" % A)
+        if not isintlike(p) or p < 0:
+            raise ValueError("non-negative integer expected as p")
+
+        super(_PowerLinearOperator, self).__init__(shape=A.shape, dtype=A.dtype, base_comm=A.base_comm)
+        self.args = (A, p)
+
+    def _power(self, fun: Callable, x: DistributedArray):
+        res = DistributedArray(global_shape=x.global_shape, dtype=x.dtype)
+        res[:] = x.local_array
+        for _ in range(self.args[1]):
+            res[:] = fun(res).local_array
+        return res
+
+    def _matvec(self, x: DistributedArray):
+        return self._power(self.args[0].matvec, x)
+
+    def _rmatvec(self, x: DistributedArray):
+        return self._power(self.args[0].rmatvec, x)
+
+
+class _ConjLinearOperator(MPILinearOperator):
+    """Complex conjugate MPI Linear Operator
+    """
+
+    def __init__(self, A: MPILinearOperator):
+        if not isinstance(A, MPILinearOperator):
+            raise TypeError('A must be a MPILinearOperator')
+        self.A = A
+        super().__init__(shape=A.shape, dtype=A.dtype, base_comm=MPI.COMM_WORLD)
+
+    def _matvec(self, x: DistributedArray):
+        x[:] = x.local_array.conj()
+        y = self.A.matvec(x)
+        if y is not None:
+            y[:] = y.local_array.conj()
+        return y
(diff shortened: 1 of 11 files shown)


pylops_mpi/basicoperators/__init__.py
```diff
@@ -0,0 +1 @@
+from .BlockDiag import MPIBlockDiag
```
mrava87, line 1: I suggest to follow the structure of pylops https://github.com/PyLops/pylops/blob/dev/pylops/basicoperators/__init__.py

pylops_mpi/__init__.py
```diff
@@ -1,5 +1,6 @@
 from .DistributedArray import DistributedArray, Partition
 from .plotting.plotting import *
+from .basicoperators import MPIBlockDiag
```
mrava87, line 3: Why not `from .basicoperators import *` so it is already the correct form for later

pylops_mpi/basicoperators/BlockDiag.py
```diff
@@ -0,0 +1,115 @@
...
+import scipy as sp
+from mpi4py import MPI
+from typing import Optional, Sequence
+
+from pylops.utils import DTypeLike
+from pylops import LinearOperator, MatrixMult
+
+from pylops_mpi.DistributedArray import local_split, Partition
+
+sp_version = sp.__version__.split(".")
```
mrava87, line 11: Since we are building a new library, I suggest we do not support early versions of spicy (like we need to do for backward compatibility in pylops). Simply use `from scipy.sparse.linalg._interface import _get_dtype` and make 1.8.0 the minimum requirement

pylops_mpi/basicoperators/BlockDiag.py
```diff
@@ -0,0 +1,115 @@
...
+        ----------
+        shape : :obj:`tuple`
+            Operator shape
+        explicit : :obj:`bool`
+            Operator contains a matrix that can be solved explicitly (``True``) or
+            not (``False``)
+
+        Notes
+        -----
+        Refer to :obj:pylops.BlockDiag for more details.
```
mrava87, line 44: It is good to refer to pylops for the general idea of the operator, but you should add some content to explain how this specific operator works in a MPI fashion. Basically, how operators are splitted across workers, how the forward and adjoint work, etc. Try to spend more time on documentation, it is as IMPORTANT as code :)

pylops_mpi/basicoperators/BlockDiag.py
```diff
@@ -0,0 +1,115 @@
...
+
+    def _matvec(self, x):
+        # Calculate x_local using mops for different ranks
+        mops_sum = np.insert(np.cumsum(self.base_comm.allgather(self.mops)), 0, 0)
+        x_local = x[mops_sum[self.rank]: mops_sum[self.rank + 1]]
+        y = []
+        for iop, oper in enumerate(self.ops):
+            y.append(oper.matvec(x_local[self.mmops[iop]:
+                                         self.mmops[iop + 1]]))
+        if y:
```
mrava87, line 101: I am not sure I understand this. In which scenario this if statement is not verified? (I guess if self.ops was empty?)

And if this is the case, what would happen to `np.hstack(self.base_comm.allgather(y))`

pylops_mpi/basicoperators/BlockDiag.py
```diff
@@ -0,0 +1,115 @@
...
+           ops : :obj:`list`
+               Linear Operators allocated at each rank.
+        """
+        local_shapes = local_split((len(ops),), self.base_comm, Partition.SCATTER, 0)
+        x = np.insert(np.cumsum(self.base_comm.allgather(local_shapes)), 0, 0)
+        return ops[x[self.rank]:x[self.rank + 1]]
+
+    def _matvec(self, x):
+        # Calculate x_local using mops for different ranks
+        mops_sum = np.insert(np.cumsum(self.base_comm.allgather(self.mops)), 0, 0)
```
mrava87, line 95: Seems like this can be done once upfront and saved into a class member?

pylops_mpi/basicoperators/BlockDiag.py
```diff
@@ -0,0 +1,115 @@
...
+        for iop, oper in enumerate(self.ops):
+            y.append(oper.matvec(x_local[self.mmops[iop]:
+                                         self.mmops[iop + 1]]))
+        if y:
+            y = np.concatenate(y)
+        return np.hstack(self.base_comm.allgather(y))
+
+    def _rmatvec(self, x):
+        # Calculate x_local using nops for different ranks
+        nops_sum = np.insert(np.cumsum(self.base_comm.allgather(self.nops)), 0, 0)
```
mrava87, line 107: Same as above

mrava87 (commented):
@rohanbabbar04, thanks for this PR. So far I struggle a bit to understand the rationale behind this operator. I had a few comments in the closed PR that I put here again, hopefully your answers (and example!) will help clarify my doubts.

- [Taken from closed PR] ALWAYS work with 1D vectors first! In your test if MPIBlockDiag you make a 3d array… when solving inverse problems we work with 1D arrays (internally operators may reshape them but from the outside - and from the solver point of view- they are 1D)… we need tests to prove that the operators we build work with 1D arrays (see tests in pylops where flatten/ravel is used in most places
- Looking at your example, I am not sure I understand how you envision/designed MPIBlockDiag… my idea is that if you have a block diagonal operator with N operators (N being the number of ranks), each rank will build one pylops operator and pass it to the MPIBlockDiag, so that once matvec/rmatvec is called each rank will compute their portion… so if you were to compare your new implementation with an equivalent pylops one, the pylops operator will have to take all of those operators and give the same result…. For simplicity take this example: you have 3 ranks and 3 5x5 matrices which you wrap into 3 MatrixMult operators. The first is filled with 1s, the second with 2s, the third with 3s.. now for pylops-MPI each rank makes one MatrixMult operator and passes it to MPIBlockDiag… the x is also distributed and can be just full of 1s for simplicity. Unless I don’t understand your example it doesn’t seem like it is built like this, rather each rank seems to make all of the ops…
- Based on the above, it seems like your `x` and `y` are not `DistributedArray`, instead just numpy arrays. Why? Again, I would think we should have distributed arrays passed to MPI operators, so that internally it is easy to get the portion needed
- In the PR you closed you had created a `LinearOperator` class and subclassed `MPIBlockDiag`. Was there any reason/benefit from doing that, or right now you do not see any use wrt `MPIBlockDiag`?

pylops_mpi/basicoperators/BlockDiag.py
```diff
@@ -0,0 +1,115 @@
...
+           ops : :obj:`list`
+               Linear operators to be stacked.
+
+           Returns
+           -------
+           ops : :obj:`list`
+               Linear Operators allocated at each rank.
+        """
+        local_shapes = local_split((len(ops),), self.base_comm, Partition.SCATTER, 0)
+        x = np.insert(np.cumsum(self.base_comm.allgather(local_shapes)), 0, 0)
```
mrava87, line 90: `x` is a bit of a unclear nomenclature, maybe you something a bit more verbose like `iops`

pylops_mpi/basicoperators/__init__.py
rohanbabbar04: Done...

pylops_mpi/__init__.py
rohanbabbar04: Done...

pylops_mpi/basicoperators/BlockDiag.py
rohanbabbar04: Updated scipy to `1.8.0`, python is also now `3.8`(as required by scipy), updated the code, also updated `requirements-dev.txt`, `environment.yml` and `environment-dev.yml`. Also removed `3.7` from build.yaml.

pylops_mpi/basicoperators/BlockDiag.py
rohanbabbar04: Updated the notes, would want your feedback...

pylops_mpi/basicoperators/BlockDiag.py
rohanbabbar04: Now it is removed  as we use DistributedArray.

pylops_mpi/basicoperators/BlockDiag.py
rohanbabbar04: Removed as not required....

pylops_mpi/basicoperators/BlockDiag.py
rohanbabbar04: Removed as not required

pylops_mpi/basicoperators/BlockDiag.py
rohanbabbar04: Removed....

pylops_mpi/basicoperators/__init__.py
mrava87: Great :)

pylops_mpi/basicoperators/BlockDiag.py
```diff
@@ -0,0 +1,139 @@
...
+            Type of elements in input array.
+
+        Attributes
+        ----------
+        shape : :obj:`tuple`
+            Operator shape
+        explicit : :obj:`bool`
+            Operator contains a matrix that can be solved explicitly (``True``) or not (``False``)
+
+        Notes
```
mrava87, line 35: Very good :D

pylops_mpi/basicoperators/BlockDiag.py
mrava87: Perfect, I don't think we need to maintain python3.7 given that pretty much everyone uses 3.8+ these days...

pylops_mpi/basicoperators/BlockDiag.py
mrava87: There is also a slight inconsistency with the fact that we allow each rank to have not just one operator but a list of operators... let's maybe simply add.

```
Each rank has one or more :class:`pylops.LinearOperator`, which we represent here compactly as :math:`\mathbf{L}_i` for rank :math:`i`.

Comment (mrava87) [reply pylops_mpi/basicoperators/BlockDiag.py:35]: Finally

**L** -> :math:`\mathbf{L}`

File: pylops_mpi/basicoperators/BlockDiag.py [hunk 10]
```diff
@@ -0,0 +1,139 @@
...
+        nops = np.zeros(len(self.ops), dtype=np.int64)
+        for iop, oper in enumerate(self.ops):
+            nops[iop] = oper.shape[0]
+            mops[iop] = oper.shape[1]
+        self.mops = mops.sum()
+        self.nops = nops.sum()
+        self.nnops = np.insert(np.cumsum(nops), 0, 0)
+        self.mmops = np.insert(np.cumsum(mops), 0, 0)
+        # Shape of the operator is equal to the sum of nops and mops at all ranks
+        self.shape = (self.base_comm.allreduce(self.nops), self.base_comm.allreduce(self.mops))
```
mrava87, line 107: you should not do self.shape, and then pass it to super... make just a temporary shape and then pass it. Same for dtype (basically the way you did clinear is how you should do all)

pylops_mpi/basicoperators/BlockDiag.py
```diff
@@ -0,0 +1,139 @@
...
+        self.localop_shape = (self.nops, self.mops)
+        if dtype:
+            self.dtype = _get_dtype(dtype)
+        else:
+            self.dtype = np.dtype(dtype)
+        clinear = all([getattr(oper, "clinear", True) for oper in self.ops])
+        super().__init__(shape=self.shape, dtype=self.dtype, clinear=clinear)
+
+    def _matvec(self, x: DistributedArray) -> DistributedArray:
+        if x.local_shape != (self.mops, ):
```
mrava87, line 118: great!

pylops_mpi/basicoperators/BlockDiag.py
```diff
@@ -0,0 +1,139 @@
...
+            raise ValueError(f"Dimension mismatch: x shape-{x.local_shape} does not match operator shape "
+                             f"{self.localop_shape}; {x.local_shape[0]} != {self.mops} (dim1) at rank={self.rank}")
+        y = DistributedArray(global_shape=self.shape[0], dtype=x.dtype)
+        y1 = []
+        for iop, oper in enumerate(self.ops):
+            y1.append(oper.matvec(x.local_array[self.mmops[iop]:
+                                                self.mmops[iop + 1]]))
+        if y1:
```
mrava87, line 126: I still dont understand this... when y1 will be empty? and if it is empty, then what y would be (I guess just some zeros as you never assign anything to it?)

pylops_mpi/basicoperators/BlockDiag.py
```diff
@@ -0,0 +1,139 @@
...
+    def _rmatvec(self, x: DistributedArray) -> DistributedArray:
+        if x.local_shape != (self.nops, ):
+            raise ValueError(f"Dimension mismatch: x shape-{x.local_shape} does not match operator shape "
+                             f"{self.localop_shape}; {x.local_shape[0]} != {self.nops} (dim0) at rank={self.rank}")
+        y = DistributedArray(global_shape=self.shape[1], dtype=x.dtype)
+        y1 = []
+        for iop, oper in enumerate(self.ops):
+            y1.append(oper.rmatvec(x.local_array[self.nnops[iop]:
+                                                 self.nnops[iop + 1]]))
+        if y1:
```
mrava87, line 139: Same as above

tests/test_blockdiag.py
```diff
@@ -0,0 +1,49 @@
...
+    assert isinstance(y_rmat, pylops_mpi.DistributedArray)
+
+    x_mat_mpi = x_mat.asarray()
+    y_rmat_mpi = y_rmat.asarray()
+
+    if rank == 0:
+        ops = [pylops.MatrixMult((i + 1) * np.ones(shape=(par['ny'], par['nx'])).astype(par['dtype'])) for i in range(size)]
+        BDiag = pylops.BlockDiag(ops=ops)
+
+        print(x_global.shape)
```
mrava87, line 45: Remove this

pylops_mpi/basicoperators/BlockDiag.py
rohanbabbar04: Done :slightly_smiling_face:

tests/test_blockdiag.py
```diff
@@ -0,0 +1,49 @@
...
+
+    x = pylops_mpi.DistributedArray(global_shape=size * par['nx'], dtype=par['dtype'])
+    x[:] = np.ones(shape=par['nx'], dtype=par['dtype'])
+    x_global = x.asarray()
+
+    y = pylops_mpi.DistributedArray(global_shape=size * par['ny'], dtype=par['dtype'])
+    y[:] = np.ones(shape=par['ny'], dtype=par['dtype'])
+    y_global = y.asarray()
+
+    x_mat = BDiag_MPI._matvec(x)
```
mrava87, line 33: I did use `_matvec` in my quick tests because I realized we could not use ``matvec / @` without making a new LinearOperator, but ideally we should include it in this PR and then never use internal methods like `_matvec ` in tests :)

rohanbabbar04:
@mrava87 Done the changes, if the code and tests are good to go, we can merge this
I will make a new PR for the example(Post Stack one)

mrava87 (commented):
Good job! Still quite a few comments but I think we are getting closer to having a good MPIBlockDiag operator.

For now I think we can stick to this implementation. Later we can consider maybe adding also what you had originally, allowing users to choose (depending on the scenario) if they want to create&pass operators per rank or all of them from all ranks - the second may be more flexible for complex scenarios but I suspect the former will be the most used one as in most cases operators will be the same for all ranks, just filled with different 'content'. (or just doing the same think to different parts of the model and data vector)

pylops_mpi/basicoperators/BlockDiag.py
```diff
@@ -0,0 +1,139 @@
...
+from pylops.utils import DTypeLike
+from pylops import LinearOperator
+
+from pylops_mpi import DistributedArray
+
+
+class MPIBlockDiag(LinearOperator):
+    r"""MPI Block-diagonal operator.
+
+        Create a block-diagonal operator from N linear operators using MPI.
```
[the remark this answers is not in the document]
rohanbabbar04: Updated the docs...

pylops_mpi/basicoperators/BlockDiag.py
```diff
@@ -0,0 +1,139 @@
...
+
+        Parameters
+        ----------
+        ops : :obj:`list`
+            Linear operators to be stacked. Alternatively,
```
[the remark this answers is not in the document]
rohanbabbar04: Changed to  One or more `class:pylops.LinearOperator`

pylops_mpi/basicoperators/BlockDiag.py
```diff
@@ -0,0 +1,139 @@
...
+            :obj:`numpy.ndarray` or :obj:`scipy.sparse` matrices can be passed
```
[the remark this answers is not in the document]
rohanbabbar04: True, have changed that. In my earlier version i had that, forgot to remove it...
Done :)

pylops_mpi/basicoperators/BlockDiag.py
```diff
@@ -0,0 +1,139 @@
...
+from mpi4py import MPI
+from typing import Optional, Sequence
+
+from pylops.utils import DTypeLike
+from pylops import LinearOperator
+
+from pylops_mpi import DistributedArray
+
+
+class MPIBlockDiag(LinearOperator):
```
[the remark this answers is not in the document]
rohanbabbar04: Changed Now a New class is created `MPILinearOperator`

pylops_mpi/basicoperators/BlockDiag.py
rohanbabbar04: Done....

pylops_mpi/basicoperators/BlockDiag.py
rohanbabbar04: Now since we are checking `x.local_shape != (self.mops, )` this should solve our problem 
Thus, removed `if y1`

pylops_mpi/basicoperators/BlockDiag.py
rohanbabbar04: Similar to above, Since we are checking x.local_shape != (self.nops, ) this should solve our problem
 Thus, removed if y1

tests/test_blockdiag.py
rohanbabbar04: Changed it from `_matvec` -> `@`
Changed it from `_rmatvec` -> `.H @`

rohanbabbar04:
Hi @mrava87, 
- I have added `MPILinearOperator` class and made `MPIBlockDiag` extend that.
- Added important classes(`_ScaledLinearOperator`,` _AdjointLinearOperator`, etc) and made them input and output `DistributedArray`.
- For the linearop tests(for Scaled, Prod and Sum) I have used `MPIBlockDiag`, as our linearop.

tests/test_linearop.py
```diff
@@ -0,0 +1,254 @@
...
+    assert isinstance(BDiag_MPI.H, MPILinearOperator)
+    assert isinstance(BDiag_MPI.T, MPILinearOperator)
+    assert isinstance(BDiag_MPI * -3, MPILinearOperator)
+    assert isinstance(BDiag_MPI.conj(), MPILinearOperator)
+    assert isinstance(BDiag_MPI + BDiag_MPI, MPILinearOperator)
+
+
+@pytest.mark.mpi(min_size=2)
+@pytest.mark.parametrize("par", [(par1), (par1j)])
+def test_square_linearop(par):
```
mrava87, line 36: Add short doctoring explaining what this test is

tests/test_linearop.py
```diff
@@ -0,0 +1,254 @@
...
+
+par1 = {'ny': 101, 'nx': 101, 'dtype': np.float64}
+par1j = {'ny': 101, 'nx': 101, 'dtype': np.complex128}
+par2 = {'ny': 301, 'nx': 101, 'dtype': np.float64}
+par2j = {'ny': 301, 'nx': 101, 'dtype': np.complex128}
+
+
+@pytest.mark.mpi(min_size=2)
+@pytest.mark.parametrize("par", [(par1), (par1j), (par2), (par2j)])
+def test_linearop(par):
```
mrava87, line 23: Add short doctoring explaining what this test is

mrava87 (commented):
This looks great!

I made just a few comments on the LinearOperator. Most should be easy to handle, but there is one or two where I cannot understand some of your code choices that would be good to discuss/agree upon before moving ahead.

After that, for me this is ready to be merged :)

tests/test_linearop.py
rohanbabbar04: Done

tests/test_linearop.py
rohanbabbar04: Done

pylops_mpi/LinearOperator.py
```diff
@@ -0,0 +1,332 @@
...
+from scipy.sparse._sputils import isintlike
+from scipy.sparse.linalg._interface import _get_dtype
+
+from pylops import LinearOperator
+from pylops.utils import DTypeLike, ShapeLike
+
+from pylops_mpi import DistributedArray
+
+
+class MPILinearOperator(LinearOperator):
```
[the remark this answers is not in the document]
rohanbabbar04: Made MPILinearOperator a standalone class

pylops_mpi/LinearOperator.py
```diff
@@ -0,0 +1,332 @@
...
+    Op : :obj:`pylops.LinearOperator`, optional
+        Linear Operator. Defaults to ``None``.
+    base_comm : :obj:`mpi4py.MPI.Comm`, optional
+        MPI Base Communicator. Defaults to ``mpi4py.MPI.COMM_WORLD``.
+
+    """
+
+    def __init__(self, shape: ShapeLike, dtype: DTypeLike, Op: LinearOperator = None,
+                 base_comm: MPI.Comm = MPI.COMM_WORLD):
+        self.Op = Op
```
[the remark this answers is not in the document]
rohanbabbar04: Since I have made a standalone class
So keeping it as `self.Op = Op`

pylops_mpi/LinearOperator.py
```diff
@@ -0,0 +1,332 @@
...
+        super().__init__(Op=Op, dtype=dtype, shape=shape)
+
+    def matvec(self, x: DistributedArray) -> DistributedArray:
+        if self.Op:
+            y = DistributedArray(global_shape=self.base_comm.allreduce(self.shape[1]), dtype=x.dtype)
+            if isinstance(x, DistributedArray):
+                x = x.local_array
+            y[:] = self.Op._matvec(x)
+        else:
+            y = self._matvec(x)
```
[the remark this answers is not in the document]
rohanbabbar04: Added `_matvec` and `_rmatvec` just like we do in pylops...

rohanbabbar04:
- I have made `MPILinearOperator` class a standalone class(not extending `pylops.LinearOperato`r), have add some docstrings comment for different methods as a reason.
- Currently dot only supports matvec operations, not matmat.
- Thinking about `asmpilinearoperator`, currently reverted from `self.base_comm.allreduce(self.shape[1])` to `self.shape[1]`, but will have to handled for different operators as a direct self.Op._matvec(x.local_array) will possibly give an error...
- For the `MPILinearOperator`, there are mainly 3 parameters we primarily use which was in `pylops.LinearOperator`, which are Op, shape and dtype..

mrava87:
Great, I think this is ready to go. We will see if things like `matmat` are needed in future but for now what we have in `MPILinearOperator` should be enough to be able to successfully solve our initial use-cases :)

PR #27, opened 2023-07-05
Added post stack inversion example

rohanbabbar04:
- Added plots for forward(data), adjoint, model and smoothmodel
- The model has a shape of (ny_i, nx, nz)

Currently the 2-D model `m = np.log((rank + 1) * np.random.uniform(1.8, 3.8, (nx, nz)))`, if this needs to be changed do let me know


mrava87:
Thanks.

Yes, the model you should use is this one https://pylops.readthedocs.io/en/stable/tutorials/poststack.html#sphx-glr-tutorials-poststack-py

Make a `testdata` folder like in pylops and copy that file. Then load the file and repeat over the y dimensions. Sorry I thought we discussed this before, but maybe I wasn't clear. Tutorial should be based on real use cases, not random numbers (those are good for tests only)

mrava87:
I will do my review after you have changed the mdoel

rohanbabbar04:
> Thanks.
> 
> Yes, the model you should use is this one https://pylops.readthedocs.io/en/stable/tutorials/poststack.html#sphx-glr-tutorials-poststack-py
> 
> Make a `testdata` folder like in pylops and copy that file. Then load the file and repeat over the y dimensions. Sorry I thought we discussed this before, but maybe I wasn't clear. Tutorial should be based on real use cases, not random numbers (those are good for tests only)

My bad....
I have loaded the model shape = (nz, nx). I have used `np.tile` on the to add stack it up so our shape will be (ny_i, nz, nx). Then with each rank we create a PostStackLinearModelling and push it to a MPIBlockDiag....

rohanbabbar04:
i will make a new PR regarding this...

mrava87:
Sure, or you could continue here if you prefer

PR #31, opened 2023-07-07
Fix servedoc rule

mrava87:
@rohanbabbar04 I noticed that using the servedoc rule I was getting an error. I realized that the path of the doc seems to be different from that of pylops (missing an html folder), so with this small change things work for me.

Please check this is also the case for you and merge if you experience the same :)


rohanbabbar04:
Thanks !
True should be only `docs/build`

Issue #33, opened 2023-07-10
MPIVStack


mrava87:
The MPIVStack operator should be implemented following these guidelines:

- [ ] operators:  one pylops operator (or a list to be consistent with what you did for BlockDiag) per rank. They must all have the same number of columns - shape[1].
- [ ] forward: take a DistributedArray in broadcast modeand return a DistributedArray in scatter mode. As the input is now the same for all operators (number of columns of the VStack operator is the same as that of every individual one), this is why we need it to be present everywhere. The output is instead distributed like that of MPIBlockDiag.
- [ ] adjoint: this needs a reduction which is the new part compared to MPIBlockDiag as all operators have outputs of the same size and they get summed into a DistributedArray operator in broadcast mode.

PR #36, opened 2023-07-12
Added MPIVStack

rohanbabbar04:
Adds #33 
- Added MPIVStack with docs.
- Added Tests

### DIFF
--- a/docs/source/api/index.rst
+++ b/docs/source/api/index.rst
@@ -40,6 +40,7 @@ Basic Operators
    :toctree: generated/
 
     MPIBlockDiag
+    MPIVStack
 
 Solvers
 -------
--- a/pylops_mpi/basicoperators/VStack.py
+++ b/pylops_mpi/basicoperators/VStack.py
@@ -108,7 +108,7 @@ class MPIVStack(MPILinearOperator):
         dtype = _get_dtype(self.ops) if dtype is None else np.dtype(dtype)
         super().__init__(shape=shape, dtype=dtype, base_comm=base_comm)
 
-    def _matvec(self, x: DistributedArray):
+    def _matvec(self, x: DistributedArray) -> DistributedArray:
         if x.partition is not Partition.BROADCAST:
             raise ValueError(f"x should have partition={Partition.BROADCAST}, {x.partition} != {Partition.BROADCAST}")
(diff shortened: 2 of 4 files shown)


rohanbabbar04:
@mrava87, so now as we were working more on DistributedArray with `partition=Broadcast`, I realized that we cannot call it `broadcast ` completely since we are not sending from one process to another, and there may be instances when we can alter the value of the DistributedArray(Broadcast) at different ranks which should not be possible..
Earlier we were just filling the array using `self.local_array[index] = value` which means I have the option to change the value at different ranks which I do not think can be called broadcast(as it means creating exact copies). This may cause problem to users.
So its better we change it to `self.local_array[index] = self.base_comm.bcast(value)`, as this would maintain that the array is broadcasted from rank-0 and all the local arrays at each rank are a copy of that.
A person can use `local_array` parameter if he/she wants to do something different to the local array at a particular rank.
What do you think??

mrava87:
@rohanbabbar04 thanks for the code, will try to review it before the end of the week. 

I completely agree with your point, when we use the DistributedArray with `partition=Broadcast` we must assume that each rank has the exact same copy of the array... so may `Replicated` or `Copied` could have been a more explicit definition of what we have instead of `Broadcast`.

I am a bit afraid of the change you made as this is a very aggressive choice happening under the hood. It may be the best way to avoid inconsistency BUT we need to clearly document this in the doctoring so users/developers know! Of course the other option would be to check if any of the `local_array` differs and raise an error - but this would call for a lot of communication overhead.. one idea could be to add a `debug` parameter to `DistributedArray` which when `True` does this check and when False bypasses it... so we let the user decide and explain what are the implications of either options (this is a bit like checking for memory allocation in debugging mode, and avoid doing so once the code is ready to get performance) 

Also, what do you mean `A person can use local_array parameter if he/she wants to do something different to the local array at a particular rank.`? Can I do 

```
x = DistributedArray(...)
x.local_array = np.ones(10)*rank
```

or would this give me an error?

mrava87:
@rohanbabbar04 this is good to go for me, if you are happy with it :)

rohanbabbar04:
Yes @mrava87, good to go from my end as well....

PR #37, opened 2023-07-12
Inconsistency in broadcast

rohanbabbar04:
- Currently we assume that the person will input equal arrays at each rank

### DIFF
--- a/pylops_mpi/DistributedArray.py
+++ b/pylops_mpi/DistributedArray.py
@@ -57,13 +57,13 @@ class DistributedArray:
     Multidimensional NumPy-like distributed arrays.
     It brings NumPy arrays to high-performance computing.
 
-    .. warning:: When setting the partition to :obj:`pylops_mpi.Partition.BROADCAST`
-        of the DistributedArray, it is crucial to be aware that any attempts to
-        make arrays different from rank to rank will be overwritten by the actions
-        of rank 0. This means that if you modify the DistributedArray on a specific rank,
-        and are using broadcast to synchronize the arrays across all ranks, the
-        modifications made by other ranks will be discarded and
-        overwritten with the value at rank 0.
+    .. warning:: When setting the partition of the DistributedArray to
+        :obj:`pylops_mpi.Partition.BROADCAST`, it is crucial to be aware
+        that any attempts to make arrays different from rank to rank will be
+        overwritten by the actions of rank 0. This means that if you modify
+        the DistributedArray on a specific rank, and are using broadcast to
+        synchronize the arrays across all ranks, the modifications made by other
+        ranks will be discarded and overwritten with the value at rank 0.
 
     Parameters
     ----------
@@ -112,6 +112,14 @@ class DistributedArray:
 
         `Partition.BROADCAST` - The value at rank-0 is broadcasted
         and is assigned to all the ranks.
+
+        Parameters
+        ----------
+        index : :obj:`int` or :obj:`slice`
+            Represents the index positions where a value needs to be assigned.
+        value : :obj:`int` or :obj:`numpy.ndarray`
+            Represents the value that will be assigned to the local array at
+            the specified index positions.
         """
         if self.partition is Partition.BROADCAST:
             self.local_array[index] = self.base_comm.bcast(value)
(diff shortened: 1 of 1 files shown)


rohanbabbar04:
So now as we were working more on DistributedArray with `partition=Broadcast`, I realized that we cannot call it `broadcast ` completely since we are not sending from one process to another, and there may be instances when we can alter the value of the DistributedArray(Broadcast) at different ranks which should not be possible..
Earlier we were just filling the array using `self.local_array[index] = value` which means I have the option to change the value at different ranks which I do not think can be called broadcast(as it means creating exact copies). This may cause problem to users.
So its better we change it to `self.local_array[index] = self.base_comm.bcast(value)`, as this would maintain that the array is broadcasted from rank-0 and all the local arrays at each rank are a copy of that.
A person can use `local_array` parameter if he/she wants to do something different to the local array at a particular rank.
What do you think??

rohanbabbar04:
@rohanbabbar04 thanks for the code, will try to review it before the end of the week. 

I completely agree with your point, when we use the DistributedArray with `partition=Broadcast` we must assume that each rank has the exact same copy of the array... so may `Replicated` or `Copied` could have been a more explicit definition of what we have instead of `Broadcast`.

I am a bit afraid of the change you made as this is a very aggressive choice happening under the hood. It may be the best way to avoid inconsistency BUT we need to clearly document this in the doctoring so users/developers know! Of course the other option would be to check if any of the `local_array` differs and raise an error - but this would call for a lot of communication overhead.. one idea could be to add a `debug` parameter to `DistributedArray` which when `True` does this check and when False bypasses it... so we let the user decide and explain what are the implications of either options (this is a bit like checking for memory allocation in debugging mode, and avoid doing so once the code is ready to get performance) 

Also, what do you mean `A person can use local_array parameter if he/she wants to do something different to the local array at a particular rank.`? Can I do 

```
x = DistributedArray(...)
x.local_array = np.ones(10)*rank
```

or would this give me an error?

mrava87:
Alright, I see. Then I agree with the first 2 points.

For the debug, what I meant is that instead of having self.base_comm.bcast(value) we could have a check 

if not self.debug:
    self.base_comm.bcast(value) # force value in rank0 to all
else:
     #write a code that checks, probably in a round robin fashion (rank1 checks rank0, rank2 checks rank1), that value for all ranks is the same… if not raises an error.

Having debug=False as default, but allowing people that are writing a code for the first time to make sure their code is not attempting to have different values in different ranks (and pylops-mpi changing this under the hood for them, unaware of this).

rohanbabbar04:
Ok i got it....
Should we implement the `debug` parameter then?

mrava87:
This is suggestion, what do you think?

For sure, whether it is added or not, in this PR we need to include some of the details of what happens in the setitem method for both users and devs to know. I suggest you add a docstring to the method and briefly explain the reason for line 101 (basically a concise version of what you wrote in your PR comment) as well as a more general statement in the main docstring of the class for users, where we warn them that with broadcast any attempt to make arrays different from rank to rank will be overwritten by what rank0 does

mrava87:
Sorry, I forgot to reply this. I think a `.. warning::` is maybe a better idea (since we also have already another note in this docstring :)

rohanbabbar04:
@mrava87, as we had discussed I have added a warning and also added docstrings to the setter method(brief but explains what we do)..

mrava87 (approved):
Very good, few suggested changes:

- the partition to :obj:`pylops_mpi.Partition.BROADCAST` of the DistributedArray -> ` the partition of the DistributedArray to :obj:`pylops_mpi.Partition.BROADCAST`
- since you add a doctoring to __setitem__ it would make sense to also have a parameters section to specify what index and value are

rohanbabbar04:
> Very good, few suggested changes:
> 
>     * the partition to :obj:`pylops_mpi.Partition.BROADCAST` of the DistributedArray -> ` the partition of the DistributedArray to :obj:`pylops_mpi.Partition.BROADCAST`
> 
>     * since you add a doctoring to **setitem** it would make sense to also have a parameters section to specify what index and value are

Added in the new commit :+1:

Issue #34, opened 2023-07-10
Poststack inversion


mrava87:
Add the following parts to the example `plot_post_stack_inversion`:

- [x] inversion using the newly developed CGLS solver. 
- [ ] inversion with regularization: ~~for now we can also show how to add regularization over the 2 axes that are not distributed. This simply means making an operator for each rank that looks like:~~

```
PPop = PoststackLinearModelling(wav, nt0=nz, spatdims=(ny_i, nx))
Top = Transpose((ny_i, nx, nz), (2, 0, 1))
DxOp = SecondDerivative(...)
DzOp = SecondDerivative(...)

eps = # reg parameter
BDiag = MPIBlockDiag(ops=[Top.H @ PPop @ Top, eps * DxOp, eps * DzOp ])
```

~~and stacking d_dist with 2 vectors of zeros of the size of the local portion of the model.~~ Use pylops_mpi.Laplacian operator.

- [x] I would also suggest including an equation at the top of the tutorial explaining how the problem can be written mathematically in a distributed fashion (including the regularizer)

rohanbabbar04:
@mrava87 for the inversion portion
will we be working on two methods
- one will be cgls(like we have done lsqr for 2d)
- And second as you mentioned inv with regularization

rohanbabbar04:
I think now we can revisit this...
With `MPISecondDerivative` now ready maybe you can add another point here for updating the example....

mrava87:
Indeed

PR #38, opened 2023-07-13
Updated Inversion Example using CGLS solver

rohanbabbar04:
#34 Part-1
- Updated the example to use inversion using CGLS
- Also reduced the model shape(reduces time in building the sphinx docs, this was also implemented in this [example](https://pylops.readthedocs.io/en/stable/tutorials/poststack.html)) 
- Added the plot for inversion...

### DIFF
--- a/pylops_mpi/__init__.py
+++ b/pylops_mpi/__init__.py
@@ -1,7 +1,12 @@
 from .DistributedArray import DistributedArray, Partition
-from .LinearOperator import MPILinearOperator, asmpilinearoperator
-from .plotting.plotting import *
+from .LinearOperator import *
 from .basicoperators import *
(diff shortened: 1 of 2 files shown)


mrava87 (approved):
I have pushed a commit where I helped you improve the text, giving some more details to the user on how to set up this problem with a Distributed mindset.

I was also playing a bit with the library to see how things behave, and so I created the example with regularization. 

Have a look at the update example and if you agree with my changes, I think we can merge this. We will come back later to do one final update, adding an example with also regularization in y (but this require building the MPISecondDerivative operator first :) ).

rohanbabbar04:
Thank You :)
This looks good, I made minor changes to the code...

PR #44, opened 2023-07-17
Added Stacking examples

rohanbabbar04:
- Added `MPIVStack` example(Vertical Stacking).
- Added `MPIHStack` example(Horizontal stacking).
- Added `MPIBlockDiag` example(For Block Diagonal example).

### DIFF
--- a/examples/plot_stacking.py
+++ b/examples/plot_stacking.py
@@ -2,8 +2,8 @@ r"""
 Stacking Operators
 ==================
 
-This example shows how to use "stacking" operators such as :py:class:`pylops_mpi.basicoperators.MPIBlockDiag` and
-:py:class:`pylops_mpi.basicoperators.MPIVStack`.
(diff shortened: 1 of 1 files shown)


rohanbabbar04:
For the examples, I have used more than 1 operator(2) at each rank, I think since we use only 1 for our tests, its better to use 2 in the example(plus the plots look better)

mrava87 (approved):
This looks very good. 

One thing (maybe better to make a separate PR), I would suggest we create an operator called HStack where we just call the VStack matvec/rmatvec in reversed order… code wise it doesn’t add much but users can use it instead of using an adjoint of VStack :) what do you think?

mrava87:
Great, this is good to go :D

rohanbabbar04:
Have changed the example of horizontal stacking with `MPIHStack` instead of `MPIVStack.H`, since it makes more sense and would be easy for users to understand...

PR #45, opened 2023-07-18
Added MPIHStack

rohanbabbar04:
- Added `MPIHStack`
- Added test for MPIHStack.

MPIHStack now works similar to the `pylops.HStack`, I have used the concept -> `HStack(ops) = VStack(ops.H).H`.
```
# In code 
self.HStack = pylops_mpi.MPIVStack(ops=hops, base_comm=base_comm, dtype=dtype).H
```
which creates the HStack and then called the matvec and rmatvec of MPIVStack.
The docs is pretty much the same as MPIVStack...

### DIFF
--- a/docs/source/api/index.rst
+++ b/docs/source/api/index.rst
@@ -41,6 +41,7 @@ Basic Operators
 
     MPIBlockDiag
     MPIVStack
+    MPIHStack
 
 Solvers
 -------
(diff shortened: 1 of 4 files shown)


rohanbabbar04:
Lets first get this one in and then i will change the example #44 and merge it...

mrava87 (approved):
Good :D

Issue #47, opened 2023-07-19
Add logic to MPILinearOperator to wrap a simple PyLops operator


mrava87:
# Problem
Currently MPILinearOperator has `_matvec\_rmatvec` that simply apply the operator `_matvec\_rmatvec` if Op is provided at initialization phase. 

As it is written right now, I expect this to work only if the input array has a SCATTER partition as `y` is created without specifying the partition (and SCATTER is the default). 

However, I envision another scenario: we have a vector x with partition.BROADCAST and we want to apply the following compound operator:

```
y = [A1,
     A2,  D x
     A3]
```
where D is a native PyLops operator acting on the entire x and returning a DistributedArray of the same shape, and the next is a MPIVStack with 3 PyLops operators (we already can handle this).

# Suggested solution
We should consider equipping MPILinearOperator with some logic for the partition of the input, such that it operates correctly in both scenarios. This would allow us to do something like this:

```
n = 100
x = DistributedArray(n, partition=Partition.BROADCAST)
x[:] = 0.
Op = FirstDerivative
MPIOp = MPILinearOperator(Op)
...
VStack = MPIVStack(..)
FullOp =  VStack @ MPIOp
y = FullOp @ x

```

PR #48, opened 2023-07-20
Added support for broadcast in `_matvec` and `_rmatvec` in MPILinearOperator

rohanbabbar04:
Adds #47 
- Added support for broadcast in `_matvec` and `_rmatvec`.
- Added a test to check the above..

### DIFF
--- a/examples/plot_mpilinop.py
+++ b/examples/plot_mpilinop.py
@@ -1,0 +1,97 @@
+r"""
+MPILinearOperator
+=================
+
+This example demonstrates the use of the :py:class:`pylops_mpi.MPILinearOperator` to wrap
+PyLops Operators. PyLops operators can be converted into :py:class:`pylops_mpi.MPILinearOperator`
+using the :py:func:`pylops_mpi.asmpilinearoperator` method. Additionally, the example showcases
+how to use these wrapped PyLops operators with other operators provided by PyLops-MPI.
+
+"""
+
+import numpy as np
+from mpi4py import MPI
+from matplotlib import pyplot as plt
+
+import pylops
+
+import pylops_mpi
+
+np.random.seed(42)
+plt.close("all")
+rank = MPI.COMM_WORLD.Get_rank()
+size = MPI.COMM_WORLD.Get_size()
+
+###############################################################################
+# Let's start by creating an instance of the :py:class:`pylops.FirstDerivative`,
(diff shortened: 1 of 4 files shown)


rohanbabbar04:
@mrava87, wasn't a really big change just needed to add `partition=x.partition`, I think broadcast support will be really helpful.
- For single pylops operator we can currently use broadcast x `MPIOp @ x`.(It just calculates at each rank).
- For complex operators if we align the shape of MPIOp and Vstack with the `x.local_array` we can also make the adjoint work with scattered x ...
-  For the `broadcast(in your example)` it works pretty well and the output returned is as expected(scattered `DistributedArray`)...

mrava87:
Wow, this is nice. I didn't realize it would have been so easy. And indeed this is going to be a very useful feature.

Just two small comment: 
- I would split the test `test_mpilinop` into 3 smaller tests, it is usually good practice not to test too many features at the same time.  
- I would also add an example, basically the same or similar code of the `# Test Product of MPIVStack and MPILinearOperator` with some explanation, to lift this feature up and make users aware of it :D

rohanbabbar04:
Done the changes that were requested
- Split the tests into 3 parts.
- Added an example of `MPILinearOperator`(Mostly similar, just added plots and some explaination)..

examples/plot_mpilinop.py
```diff
@@ -0,0 +1,97 @@
...
+###############################################################################
+# Now, to carry out the matrix-vector product using the MPILinearOperator, we first
+# create a :py:class:`pylops_mpi.DistributedArray` with the partition set to
+# ``pylops_mpi.Partition.BROADCAST``, denoted by :math:`x`. The matrix-vector product
+# is then computed at each rank, and the result returned is a :py:class:`pylops_mpi.DistributedArray`
+# with the same partitioning.
+x = pylops_mpi.DistributedArray(global_shape=Ny * Nx, partition=pylops_mpi.Partition.BROADCAST)
+x[:] = 1
+y = Mop @ x
+print(y)
```
mrava87, line 43: Maybe print(f'y: {y}')

mrava87 (approved):
Very good, I just made a minor comment on the example. Other than that it is good to go for me :)

examples/plot_mpilinop.py
rohanbabbar04: Done.

PR #55, opened 2023-07-28
Added MPIFirstDerivative

rohanbabbar04:
Taking into consideration #49 
- Added `MPIFirstDerivative`
- Added tests and updated docs..

### DIFF
--- a/pylops_mpi/basicoperators/FirstDerivative.py
+++ b/pylops_mpi/basicoperators/FirstDerivative.py
@@ -11,7 +11,7 @@ from pylops_mpi import (
     Partition
 )
 
-from pylops_mpi.utils.decorators import redistribute
+from pylops_mpi.utils.decorators import reshaped
 
 
 class MPIFirstDerivative(MPILinearOperator):
@@ -134,7 +134,7 @@ class MPIFirstDerivative(MPILinearOperator):
             x = DistributedArray.to_dist(x=x.local_array)
         return self._hrmatvec(x)
 
-    @redistribute
+    @reshaped
     def _matvec_forward(self, x: DistributedArray) -> DistributedArray:
         y = DistributedArray(global_shape=x.global_shape, dtype=self.dtype, axis=x.axis)
         ghosted_x = x.add_ghost_cells(cells_back=1)
@@ -144,7 +144,7 @@ class MPIFirstDerivative(MPILinearOperator):
         y[:] = y_forward / self.sampling
         return y
 
-    @redistribute
+    @reshaped
     def _rmatvec_forward(self, x: DistributedArray) -> DistributedArray:
         y = DistributedArray(global_shape=x.global_shape, dtype=self.dtype, axis=x.axis)
         y[:] = 0
@@ -160,7 +160,7 @@ class MPIFirstDerivative(MPILinearOperator):
         y[:] /= self.sampling
         return y
 
-    @redistribute
+    @reshaped
     def _matvec_backward(self, x: DistributedArray) -> DistributedArray:
         y = DistributedArray(global_shape=x.global_shape, dtype=self.dtype, axis=x.axis)
         ghosted_x = x.add_ghost_cells(cells_front=1)
@@ -170,7 +170,7 @@ class MPIFirstDerivative(MPILinearOperator):
         y[:] = y_backward / self.sampling
         return y
 
-    @redistribute
+    @reshaped
     def _rmatvec_backward(self, x: DistributedArray) -> DistributedArray:
         y = DistributedArray(global_shape=x.global_shape, dtype=self.dtype, axis=x.axis)
         y[:] = 0
@@ -186,7 +186,7 @@ class MPIFirstDerivative(MPILinearOperator):
         y[:] /= self.sampling
         return y
 
-    @redistribute
+    @reshaped
     def _matvec_centered3(self, x: DistributedArray) -> DistributedArray:
         y = DistributedArray(global_shape=x.global_shape, dtype=self.dtype, axis=x.axis)
         ghosted_x = x.add_ghost_cells(cells_front=1, cells_back=1)
@@ -204,7 +204,7 @@ class MPIFirstDerivative(MPILinearOperator):
         y[:] /= self.sampling
         return y
 
-    @redistribute
+    @reshaped
     def _rmatvec_centered3(self, x: DistributedArray) -> DistributedArray:
         y = DistributedArray(global_shape=x.global_shape, dtype=self.dtype, axis=x.axis)
         y[:] = 0
@@ -229,7 +229,7 @@ class MPIFirstDerivative(MPILinearOperator):
         y[:] /= self.sampling
         return y
 
-    @redistribute
+    @reshaped
     def _matvec_centered5(self, x: DistributedArray) -> DistributedArray:
         y = DistributedArray(global_shape=x.global_shape, dtype=self.dtype, axis=x.axis)
         ghosted_x = x.add_ghost_cells(cells_front=2, cells_back=2)
@@ -254,7 +254,7 @@ class MPIFirstDerivative(MPILinearOperator):
         y[:] /= self.sampling
         return y
 
-    @redistribute
+    @reshaped
     def _rmatvec_centered5(self, x: DistributedArray) -> DistributedArray:
         y = DistributedArray(global_shape=x.global_shape, dtype=self.dtype, axis=x.axis)
         y[:] = 0
--- a/pylops_mpi/basicoperators/FirstDerivative.py
+++ b/pylops_mpi/basicoperators/FirstDerivative.py
@@ -1,0 +1,285 @@
+from typing import Callable
+import numpy as np
+
+from pylops.utils import DTypeLike
+
+from pylops_mpi import (
+    DistributedArray,
+    MPILinearOperator,
+    Partition
+)
+
+
+class MPIFirstDerivative(MPILinearOperator):
+    r"""MPI First Derivative
+
+    Apply a first derivative using a multiple-point stencil finite-difference
+    approximation with :class:`pylops_mpi.DistributedArray`.
+
+    Parameters
+    ----------
+    dims : :obj:`int`
+        Number of samples for model and data vector.
+    sampling : :obj:`float`, optional
+        Sampling step :math:`\Delta x`.
+    kind : :obj:`str`, optional
+        Derivative kind (``forward``, ``centered``, or ``backward``).
+    edge : :obj:`bool`, optional
+        Use reduced order derivative at edges (``True``) or
+        ignore them (``False``). This is currently only available
+        for centered derivative.
+    order : :obj:`int`, optional
+        Derivative order (``3`` or ``5``). This is currently only available
+        for centered derivative.
+    dtype : :obj:`str`, optional
+        Type of elements in the input array.
+
+    Attributes
+    ----------
+    shape : :obj:`tuple`
+        Operator shape
+
+    Notes
+    -----
+    The MPIFirstDerivative operator applies a first derivative to a :class:`pylops_mpi.DistributedArray`
+    using either a first-order backward and forward stencil or a second or third ordered centered stencil.
+
+    When computing the first derivative using a :class:`pylops_mpi.DistributedArray`, a technique named **ghosting**
+    is employed, often in conjunction with MPI for communication between different processes. Ghosting involves
+    creating additional **"ghost cells"** around the boundary of each process's local data domain. These ghost cells
+    are replicated copies of the border cells from neighboring processes, these cells allow each process
+    to perform local computations without the need for explicit communication with other processes.
+
+    Now, for a one-dimensional DistributedArray, the first order forward stencil is:
+
+    .. math::
+        y[i] = (x[i+1] - x[i]) / \Delta x
+
+    while the first order backward stencil is:
+
+    .. math::
+        y[i] = (x[i] - x[i-1]) / \Delta x
+
+    and the second-order centered stencil is:
+
+    .. math::
+        y[i] = 0.5 * (x[i+1] - x[i-1]) / \Delta x
+
+    where :math:`y` is a DistributedArray and :math:`x` is a Ghosted array created by adding
+    border cells of neighboring processes to the local array at each rank.
+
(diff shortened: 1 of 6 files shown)


pylops_mpi/basicoperators/FirstDerivative.py
```diff
@@ -0,0 +1,285 @@
...
+            self._hrmatvec = self._rmatvec_backward
+        else:
+            raise NotImplementedError(
+                "'kind' must be 'forward', 'centered', or 'backward'"
+            )
+
+    def _matvec(self, x: DistributedArray) -> DistributedArray:
+        # If Partition.BROADCAST, then convert to Partition.SCATTER
+        if x.partition is Partition.BROADCAST:
+            x = DistributedArray.to_dist(x=x.local_array, axis=x.axis)
```
mrava87, line 123: mmh if the array is of `Partition.BROADCAST`, what does `x.axis` really mean? I guess from how DistributedArray is implemented right now, one could choose any axis, but here this would lead to problems... see my general comment

pylops_mpi/basicoperators/FirstDerivative.py
```diff
@@ -0,0 +1,285 @@
...
+
+    def _rmatvec(self, x: DistributedArray) -> DistributedArray:
+        # If Partition.BROADCAST, then convert to Partition.SCATTER
+        if x.partition is Partition.BROADCAST:
+            x = DistributedArray.to_dist(x=x.local_array, axis=x.axis)
+        return self._hrmatvec(x)
+
+    def _matvec_forward(self, x: DistributedArray) -> DistributedArray:
+        y = DistributedArray(global_shape=x.global_shape, dtype=self.dtype, axis=x.axis)
+        y[:] = 0
```
mrava87, line 134: is this really needed... if I look at 138, it seems to me you are setting the entire array, so we may be able to avoid setting it to zero first

mrava87 (commented):
Thanks!

I think this PR looks very good and aligned with our discussion. 

There is only one thing left, which I think we need to decide on and handle / make explicit in the doc: currently we do not provide `axis` to this operator, meaning that we implicitly assume the derivative is always applied to a given axis (looking at the code and how you write the stencils, this looks to me like it is axis=0, see for example: `y_forward = ghosted_x[1:] - ghosted_x[:-1]`. However, when you do the appends you seem to do them along `y.axis` (which is `x.axis`). I suspect that if `x.axis>0` we would not be doing the correct thing... do you agree?

So one option is to raise errors if `axis!=0`, whilst the other option is to add `axis`, check that `self.axis==x.axis` and handle the stencils over any axis. In fact if x.axis=0 but self.axis=1, it won't really make sense to use this operator as the derivative stencils do not go across ranks, and users can just leverage pylops.FirstDerivative together with pylops_mpi.BlockDiag like we have done before :)

pylops_mpi/basicoperators/FirstDerivative.py
```diff
@@ -0,0 +1,285 @@
...
+    Notes
+    -----
+    The MPIFirstDerivative operator applies a first derivative to a :class:`pylops_mpi.DistributedArray`
+    using either a first-order backward and forward stencil or a second or third ordered centered stencil.
+
+    When computing the first derivative using a :class:`pylops_mpi.DistributedArray`, a technique named **ghosting**
+    is employed, often in conjunction with MPI for communication between different processes. Ghosting involves
+    creating additional **"ghost cells"** around the boundary of each process's local data domain. These ghost cells
+    are replicated copies of the border cells from neighboring processes, these cells allow each process
+    to perform local computations without the need for explicit communication with other processes.
```
mrava87, line 51: We should also add a short comment that we currently do not handle edges, so this operator behaves like pylops equivalent with `edge=False`

tests/test_derivative.py
```diff
@@ -0,0 +1,229 @@
...
+    "dtype": np.float128,
+    "partition": pylops_mpi.Partition.SCATTER
+}
+
+
+@pytest.mark.mpi(min_size=2)
+@pytest.mark.parametrize("par", [(par1), (par1b), (par1j), (par1e), (par2), (par2b),
+                                 (par2j), (par2e), (par3), (par3b), (par3j), (par3e),
+                                 (par4), (par4b), (par4j), (par4e)])
+def test_first_derivative_forward(par):
```
mrava87, line 146: add short doctoring

tests/test_derivative.py
```diff
@@ -0,0 +1,229 @@
...
+        y_adj = y_adj_dist.asarray()
+        if rank == 0:
+            Fop = pylops.FirstDerivative(dims=par['nz'], sampling=par['dz'],
+                                         kind="centered", edge=par['edge'],
+                                         order=order, dtype=par['dtype'])
+            assert Fop_MPI.shape == Fop.shape
+            y_np = Fop @ x_global
+            y_adj_np = Fop.H @ x_global
+            assert_allclose(y, y_np, rtol=1e-14)
+            assert_allclose(y_adj, y_adj_np, rtol=1e-14)
```
mrava87, line 229: We may want to also add a test (just for one of the kind) where the input array is Nd (say 2d)... I think, this is where the problem I mentioned in the general comment will show up if you were to have x as a DistributedArray with axis!=0... tests are a useful way to catch issues, not only to check correctness.

Of course, the test to add will be with axis=0, but I guess you get my point ;)

rohanbabbar04:
> Thanks!
> 
> I think this PR looks very good and aligned with our discussion.
> 
> There is only one thing left, which I think we need to decide on and handle / make explicit in the doc: currently we do not provide `axis` to this operator, meaning that we implicitly assume the derivative is always applied to a given axis (looking at the code and how you write the stencils, this looks to me like it is axis=0, see for example: `y_forward = ghosted_x[1:] - ghosted_x[:-1]`. However, when you do the appends you seem to do them along `y.axis` (which is `x.axis`). I suspect that if `x.axis>0` we would not be doing the correct thing... do you agree?
> 
> So one option is to raise errors if `axis!=0`, whilst the other option is to add `axis`, check that `self.axis==x.axis` and handle the stencils over any axis. In fact if x.axis=0 but self.axis=1, it won't really make sense to use this operator as the derivative stencils do not go across ranks, and users can just leverage pylops.FirstDerivative together with pylops_mpi.BlockDiag like we have done before :)

Thanks
When it comes to adding a parameter `axis` to the class and calculating the stencil over that axis, since we use dims as an integer the only possible axis if I am not wrong will be -1 and 0. So when will we be encountering a self.axis =1 ?

mrava87:
Mmh my bad, I had forgotten about that... but I think we need to change it and allow Nd arrays. Let me explain why: imagine you have a 3d array `ny x nx x nz` which is distributed over `axis=0`. This is for example the case of our Poststack tutorial. Now, if you use `dims=ny*nx*nz` (since this is the only option so far) you will not be able to apply a first derivative over the y axis, which is what this operator is intended for. 

This is why I suggested to have no axis and implicitly compute the derivative over axis=0, but to handle the dimensions

rohanbabbar04:
Thanks Understood. 
I will get onto it..

mrava87:
> Thanks Understood.
> I will get onto it..

Great! Sorry, maybe we didn't think so deeply about the implications of having only 1d, but I think having a start with 1d only and now move to nd is still a probably better/easier way to do then if we immediately started with nd... and I feel there is little to be done to make this work :)

rohanbabbar04:
I think what we have done will greatly help as now we would prioritize calculating the derivative along axis=0.
The concept should remain the same, would discuss if I find any other challenges in its implementation..

rohanbabbar04:
Hi @mrava87, I think we need to go little different than the 1-D approach, I think pretty much the code in _matvec and _rmatvec will remain the same.
But to work with ND-Arrays, like we do in `pylops.FirstDerivative`, we may have to redistribute the DistributedArray, meaning we will convert the 1-D array to an ND-Array, and distribute over that axis. For converting if the `np.prod(local_shape)` is same, we can reshape it as it is. For unequal local shapes product we can use the ghost cells to make our ND DistributedArray, then we can proceed with what we have done...
What do you think @mrava87 ?

rohanbabbar04:
Hi @mrava87, I have handled the case for `dims=(...)`,some of the changes I made as doing that with only a 1-D DistributedArray was complicated..
- Added `ravel` as a method in DistributedArray class.
- Added `redistribute` decorator which reshapes the `x` to an n-d distributed array, also ravels the y.
- - The redistribute uses the ghost cells to reshape itself..
- Used the decorator on each `_matvec`,` _rmatvec`(similar to what we do for pylops.FirstDerivative), Made changes to the edge cases of 0s accordingly.
- Currently have made if work for axis=0, which currently is the main requirement(we also may be able to do for other axis, but will keep that for later, so we don't have `axis` as a parameter)
- Updated the tests wherein we take a dims as both an int and a tuple.

This is what I came up with, which seemed to be the best and apt approach what do you think?

mrava87:
Thanks @rohanbabbar04, but I am a bit confused by the approach you took... and a bit worried that we are doing a lot of additional communications that may not be needed (mostly referring to the @redistribute decorator).

First of all, I am not really sure I understand what `redistribute` does?

Second, perhaps I am missing something, but let me try to explain what I think is a easier way which does not require any communication other than for the ghost cells within the derivative part:

- you have `arr = DistributedArray(global_shape=ny*nx*nz, axis=0)`. Assuming ny is divisible by `size` then each rank has a portion of this 1d array. However, we can reinterpret this as if it was a 3d array (ny, nx, nz) and there each rank will have an array of size(ny//size, nx, nz).
- you make an operator `Fop = pylops_mpi.basicoperators.FirstDerivative(dims=(ny, nx, nz), axis=0)`
- you apply `Fop @ x`; each rank has internally access to a portion of arr or size  `(ny//size*nx*nz)` which using the knowledge of `dims` can be reshaped into `(ny//size, nx, nz)`, which can reshape (instead of redistribute you may think of having something like the pylops reshape). Now the `.add_ghost_cells` method is called and the stencil is applied.

Let's try to agree on this first before changing the code further :D

mrava87:
Ok, perhaps I get the point... are you trying to handle a case where the operator and x are not agreeing on how the array is distributed? 

I would say this should be something users have to ensure, and our codes should only check and raise an error otherwise... my worry is that if we silently do this we will end up with unefficients codes as users do not bother about aligning things and internally we do a lot more communications than needed to fix laziness from the user side... or maybe I don't get what the code is trying to handle 😕

rohanbabbar04:
Sorry I now see that the communication can be really ugly at times, so we should ask the users to reshape their arrays so that it doesn't affect other things...

rohanbabbar04:
So @mrava87 , As you mentioned, now I have raised an error if the `np.prod(dims.shape) != np.prod(x.shape)`, which means the user should change the number of processes(something divisible with dims[0])...
Other than that all your suggested points I think are covered(currently it works for `axis=0`)

mrava87:
Right, I see the point about using dims or  shape for nd or 1D arrays, which also when they are of same size, can be distributed differently… but, would this case ever happen. Here you have an array and an operator having different total size so they won’t talk to each other properly. I 

I am myself unsure right now what is best, just a few thought:

- your approach: solves the problem under the hood, but it does it in a rather expensive way… however, it is good that is things are aligned no communication happen (I didn’t realize this would be the case), so the cost of the decorator is only local (some local copies..)
- other approach: forces the user to think more, but it may be a bit stiff. Imagine you have the case you describe

rohanbabbar04:
For the other approach, one case is if the dims[0] is completely divisible by the number of processes, the first derivative can calculate.....
So I think the user can change the number of processes and make sure that it is divisible with dims[0] so we don't need to do any communication 
What do you think?

rohanbabbar04:
Also for my approach, it expensive only for the cases where there is difference in local shapes but no communication takes place when they are aligned(it is I will say more user convenient)...

mrava87:
Other that this, I think this new version looks good and almost ready to go :D

mrava87:
```
So I think the user can change the number of processes and make sure that it is divisible with dims[0] so we don't need to do any communication
```

As I said, this isn't always possible... what if this is  dims[0]=79, there is no way to find a rank that divides it... 

Also, take a look at my last comment, which I just added. Perhaps having this option + your method will be the best way to go. So lazy users will not need to think but incur in some extra communication cost (this could be also during quick development), whilst users that care about real performance will choose the split of the array in a manner that is consistent with that of the Operator...

rohanbabbar04:
```
The only way I can think of that strikes a balance between performance and user friendliness is to allow users to choose the splitting of the array, by adding a parameter to DistributedArray. We can default it to None and leave the current behavior as default, and let users choose it only in scenarios where this is useful.
```
Can you explain it a bit more?
Will we allow users to decide the local_shapes at each rank???

mrava87:
That's correct. So far the local_shape is automatically detected, which may be what we want to do in most scenarios, but in few cases like this one we will have more freedom to choose a different subdivision between ranks such that things align with an operator acting on one axis of a reshaped view of the total array :)

I think this should be doable without modifying DistributedArray that much, isn't it?

rohanbabbar04:
So if I am right we will just ask them the local shapes
Like `DistributedArray(500, local_shapes=[(200,), (200,), (100,)]`

mrava87:
Indeed, with default local_shape=None.. and build it from the global_shape as we do already

rohanbabbar04:
Okay... to summarize everything 
- We give an option to the users to select local shapes to reduce communication.
- We also provide a way to users for flexibility if he/she puts local shapes to None, using the more expensive method.
- And for the ones which align well with the dims local shape we solve it as it is...

rohanbabbar04:
> Indeed, with default local_shape=None.. and build it from the global_shape as we do already

Yeah changes shouldn't be much..

rohanbabbar04:
Hi @mrava87, like we discussed this PR contains and solves 2 problems:
- Dims aligned(Solved)
- Dims not aligned(we add more communication to align)

I will add the `local_shapes` paramater in a new PR, as I think we may want to handle it whenever in our codebase DistributedArray is called which is easy as we just need to add parameter everywhere, so I think we can add it in a new PR..
Plus I was thinking we should add it after doing the MPISecondDerivative, as the concept remains the same...

mrava87:
Perfect, agree!

rohanbabbar04:
> Perfect, agree!

Is this good to go?

mrava87:
Sure :)

PR #57, opened 2023-07-31
Updated tests in test_solver

### LINKED ISSUE #51: Incorrect inversion results in test_solver
I was looking at the tests in `test_solver` and I noticed that all asserts are done between the inverted vector of the pylops solver and its pylops-mpi equivalent. 

In general, I agree this is the correct way of doing so, as we trust the pylops solvers and provided we get same results we are happy with the mpi implementation. However, I noticed that the way the BlockDiag matrix is created in the first test leads to an ill-posed problem and the solution that we get is very far from the true solution. I think we should try to change this and have a problem that can converge to the true solution so that we can have a check in place that our solver does not diverge after some iterations and follows the trajectory of pylops solver all the way.

A suggestion is to make the matrix like this:

```
mats = [np.random.normal(0, 1, (par["ny"], par["nx"])) + par[
    "imag"] * np.random.normal(0, 1, (par["ny"], par["nx"])) for i in range(size)]
ops = [MatrixMult(np.conj(mats[i].T) @ mats[i] + 1e-5 *  np.eye(par["nx"]), dtype=par['dtype']) for i in range(size)]
BDiag = BlockDiag(ops=ops, forceflat=True)
```

which guarantees BDiag to be positive definite matrix and so fully invertible.

For the overdetermined (VStack) you can do the same, whilst for the underdetermined (HStack) case we cannot really recover the full solution so I would say we can keep what we have

rohanbabbar04:
Closes #51 
- Updated tests for MPIBlockDiag and MPIVStack

In the example, you used `BlockDiag(..., forceflat=True)`, but I think `forceflat` has not been released yet, so kept it as BlockDiag.

### DIFF
--- a/tests/test_solver.py
+++ b/tests/test_solver.py
@@ -86,7 +86,7 @@ par4j = {
     "par", [(par1), (par1j), (par2), (par2j), (par3), (par3j), (par4), (par4j)]
 )
 def test_cgls(par):
-    """CGLS with MPILinearOperator"""
+    """CGLS with MPIBlockDiag"""
     A = np.ones((par["ny"], par["nx"])) + par[
         "imag"] * np.ones((par["ny"], par["nx"]))
     Aop = MatrixMult(np.conj(A.T) @ A + 1e-5 * np.eye(par["nx"], dtype=par['dtype']),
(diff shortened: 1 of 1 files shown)


mrava87:
Ops yes, this is not really needed here. Looks good :)

PR #68, opened 2023-08-16
Added MPILaplacian to reg inversion example

### LINKED ISSUE #34: Poststack inversion

rohanbabbar04:
Closes #34 
- Updated example for regularized inversion to use `MPILaplacian`

### DIFF
--- a/tutorials/plot_post_stack_inversion.py
+++ b/tutorials/plot_post_stack_inversion.py
@@ -133,20 +133,47 @@ d_0 = d_dist.asarray().reshape((ny, nx, nz))
 #
 # * Inversion calculated iteratively using the :py:class:`pylops_mpi.optimization.cls_basic.CGLS` solver.
 #
-# * Inversion with spatial regularization along all three dimensions (x, y and z).
-#   This requires extending the operator in the following manner:
+# * Inversion with spatial regularization using normal equations along all three dimensions (x, y and z).
+#   This requires extending the operator and data in the following manner:
 #
 #  .. math::
-#    \mathbf{RegOp} =
+#    \mathbf{NormEqOp} =
 #    \begin{bmatrix}
 #         \mathbf{G}_1  & \mathbf{0}   &  \ldots &  \mathbf{0}  \\
 #         \mathbf{0}    & \mathbf{G}_2 &  \ldots &  \mathbf{0}  \\
 #         \vdots        & \vdots       &  \ddots &  \vdots         \\
 #         \mathbf{0}    & \mathbf{0}   &  \ldots &  \mathbf{G}_N
-#    \end{bmatrix} + \epsilon \mathbf{LapOp}
+#    \end{bmatrix}
+#    \begin{bmatrix}
+#         \mathbf{G}_1^H  & \mathbf{0}   &  \ldots &  \mathbf{0}  \\
+#         \mathbf{0}    & \mathbf{G}_2^H &  \ldots &  \mathbf{0}  \\
+#         \vdots        & \vdots       &  \ddots &  \vdots         \\
+#         \mathbf{0}    & \mathbf{0}   &  \ldots &  \mathbf{G}_N^H
+#    \end{bmatrix} + \epsilon \mathbf{LapOp} \\
+#
+#  .. math::
+#   \begin{bmatrix}
+#         \mathbf{d}_{1}^{Norm}  \\
+#         \mathbf{d}_{2}^{Norm}  \\
+#         \vdots     \\
+#         \mathbf{d}_{N}^{Norm}
+#    \end{bmatrix} =
+#    \begin{bmatrix}
+#         \mathbf{G}_1^H  & \mathbf{0}   &  \ldots &  \mathbf{0}  \\
+#         \mathbf{0}    & \mathbf{G}_2^H &  \ldots &  \mathbf{0}  \\
+#         \vdots        & \vdots       &  \ddots &  \vdots         \\
+#         \mathbf{0}    & \mathbf{0}   &  \ldots &  \mathbf{G}_N^H
+#    \end{bmatrix}
+#    \begin{bmatrix}
+#         \mathbf{d}_{1}  \\
+#         \mathbf{d}_{2}  \\
+#         \vdots     \\
+#         \mathbf{d}_{N}
+#    \end{bmatrix}
 #
 # where :math:`\mathbf{LapOp}` is the :py:class:`pylops_mpi.basicoperators.MPILaplacian` operator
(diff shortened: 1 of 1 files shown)


mrava87 (changes requested):
@rohanbabbar04 thanks for the PR.

I am not sure I agree with what you have done here though. When you want to solve an inverse problem with a Laplacian regularizer you need to write an augmented operator

[G.        m =[d,
eps*L].        0]

whose normal equations are: G^TG+eps*2L^TL = G^Td

whilst you write (G+eps*L)=d whose normal equations are G^TG+eps*2L^TL + 2eps(G^TL) = G^Td + eps*L^Td... 2eps(G^TL) and eps*L^Td should not be there.

So what you coded is wrong from a mathematical point of view. Now, I didn't think too much myself about this, I assumed once we have Laplacian we can just use it, but I think we need to find a smart way to be able to have in VStack both pylops and pylops-mpi operator.

I'm also going to think about this and we can use this PR to brainstorm until we have a working idea :)

mrava87:
I have been thinking a bit, but I am not sure yet about what is the best way forward here. 

Two possible directions that came to my mind, the first is probably easier and the best one to start, the second may be more flexible but also a bit more cumbersome under-the-hood:

1. Show in the tutorial how to solve the normal equations, basically:
```
# Regularized inversion with normal equations
epsR = 1e1
LOp = pylops_mpi.MPILaplacian(dims=(ny, nx, nz), axes=(0, 1, 2),
                              weights=(1, 1, 1), sampling=(1, 1, 1))
NormEqOp = BDiag.H @ BDiag + epsR * LOp
dnorm_dist = BDiag.H @ d_dist
minv3d_reg_dist = pylops_mpi.optimization.basic.cgls(NormEqOp, dnorm_dist, x0=mback3d_dist, niter=100, show=True)[0]
```
For now I am using `pylops_mpi.optimization.basic.cgls` but since `NormEqOp` is by definition square and semi-def positive, the most appropriate way to solve it would be via `cg`. I suggest we also add `cg`: since you did all the hard work with `cgls`, it should be just a matter of repeating the same process using pylops` cg as starting point.

2. Create two new entities: StackedDistributedArray, an object that can take multiple DistributedArray and 'stack' them on top of each other. This will be only able to interact with the other new entity `XXVStack` (not sure about the name yet...) that stacks `pylops_mpi` operators (note this is different from our `MPIVStack` where we stack pylops operators), such that when matvec is applied, the different operators act on the input DistributedArray and create an output StackedDistributedArray. Conversely, in rmatvec, the different operators act on the individual DistributedArray in the imputStackedDistributedArray and return a DistributedArray with the sum of the results. Of course, something similar could be done with `XXBlockDiag`... I am not sure about this yet, as it seems to add an extra layer of operators but so far this is the only way I can think of to solve the problem without having to sacrifice and say we can only solve normal equations...

@rohanbabbar04, any idea from your side?

rohanbabbar04:
- I think the first one makes sense to me. I will begin working on it and create a new PR for the CG Solver. Once that's done, we can jump back and finish this PR.
- Regarding the second option, I will take some time to think about it but your approach does seems reasonable.

mrava87:
Perfect :) good plan!

rohanbabbar04:
> I have been thinking a bit, but I am not sure yet about what is the best way forward here.
> 
> Two possible directions that came to my mind, the first is probably easier and the best one to start, the second may be more flexible but also a bit more cumbersome under-the-hood:
> 
>     1. Show in the tutorial how to solve the normal equations, basically:
> 
> 
> ```
> # Regularized inversion with normal equations
> epsR = 1e0 
> LOp = pylops_mpi.MPILaplacian(dims=(ny, nx, nz), axes=(0, 1, 2),
>                               weights=(1, 1, 1), sampling=(1, 1, 1))
> NormEqOp = BDiag.H @ BDiag + epsR * LOp
> dnorm_dist = BDiag.H @ d_dist
> minv3d_reg_dist = pylops_mpi.optimization.basic.cgls(NormEqOp, dnorm_dist, x0=mback3d_dist, niter=100, show=True)[0]
> ```
> 
> where `epsR = 1e0` is just chosen randomly, not sure if this is best... For now I am using `pylops_mpi.optimization.basic.cgls` but since `NormEqOp` is by definition square and semi-def positive, the most appropriate way to solve it would be via `cg`. I suggest we also add `cg`: since you did all the hard work with `cgls`, it should be just a matter of repeating the same process using pylops` cg as starting point.
> 
>     2. Create two new entities: StackedDistributedArray, an object that can take multiple DistributedArray and 'stack' them on top of each other. This will be only able to interact with the other new entity `XXVStack` (not sure about the name yet...) that stacks `pylops_mpi` operators (note this is different from our `MPIVStack` where we stack pylops operators), such that when matvec is applied, the different operators act on the input DistributedArray and create an output StackedDistributedArray. Conversely, in rmatvec, the different operators act on the individual DistributedArray in the imputStackedDistributedArray and return a DistributedArray with the sum of the results. Of course, something similar could be done with `XXBlockDiag`... I am not sure about this yet, as it seems to add an extra layer of operators but so far this is the only way I can think of to solve the problem without having to sacrifice and say we can only solve normal equations...
> 
> 
> @rohanbabbar04, any idea from your side?

I think with the `CG` now done we can get back to this.
We can pin the second point as an issue so that we don't forget and can discuss there.
For the first point,any changes which needs to be done to accomodate `cg`?

mrava87:
Agree. For the first point, I think if you use the code I pasted above and change the solver you should obtain correct results - of course, to be sure you can always check this solving the pylops equivalent problem :)

rohanbabbar04:
@mrava87 I have updated the example...

mrava87 (approved):
Thanks, looks good to me :)

PR #73, opened 2023-08-20
Added CG Solver

rohanbabbar04:
- Added `CG` class
- Added `cg` method
- Added test for cg
- Added docs


mrava87 (approved):
Great, this looks good to me :)

PR #74, opened 2023-08-20
Added deploy gh-action

mrava87:
This PR adds the Github Action used to deploy pylops-mpi on PyPI every time a release is created

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -1,0 +1,28 @@
+# This workflow uploads PyLops-MPI on PyPI using Twine when a release is created
+# For more information see: https://help.github.com/en/actions/language-and-framework-guides/using-python-with-github-actions#publishing-to-package-registries
+name: PyLopsMPI-deploy
+
+on:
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,28 @@
...
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
```
rohanbabbar04, line 13: I think we can change this to `actions/checkout@v3`

.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,28 @@
...
+      - name: Set up Python
+        uses: actions/setup-python@v2
```
rohanbabbar04, line 15: And this to `actions/setup-python@v4` as this is the latest version

mrava87:
Sounds good, feel free to go ahead and change :)

Issue #88, opened 2024-03-08
Tests for Cg and Cgls withtackDistributedArray


mrava87:
In https://github.com/PyLops/pylops-mpi/pull/86/, the `cg` and `cgls` solvers have been modified to handle `DistributedArray` and `StackDistributedArray` interchangeably.

We need to add tests for these two solvers when using `StackDistributedArray`.

PR #90, opened 2024-04-02
CG/CGLS supporting StackedDistributedArray

### LINKED ISSUE #87: Cg and Cgls with DistributedArray/StackDistributedArray
In https://github.com/PyLops/pylops-mpi/pull/86/, the `cg` and `cgls` solvers have been modified to handle `DistributedArray` and `StackDistributedArray` interchangeably.

This change is however sub-optimal in some places where in-place operations have been replaced by operations that re-initialize a `DistributedArray` or `StackDistributedArray`, eg:

```
self.c[:] = self.r.local_array + b * self.c.local_array
```

to
```
self.c = self.r + b * self.c
```

We should decide whether this is the best way or another approach could be taken to avoid re-initialization without having to create branched codes for the different type of arrays

rohanbabbar04:
Closes #87 and #88 

- Add `base_comm` as a parameter to StackedDistributedArray(Required for cg/cgls, plus it is in all our ops, array)
- Removed code for `x0`, handle printing of steps/solvers.
- Change all the docstrings to support MPIStackedLinearOperator.
- In `lsm` and `plot_cgls`, handle initial guess `x0`.
- Update the tests for existing cg/cgls tests.
- Add new tests for StackedLinearOperators and StackedDistributedArray.

### DIFF
--- a/examples/plot_cgls.py
+++ b/examples/plot_cgls.py
@@ -45,7 +45,11 @@ y = BDiag @ x
 # are then obtained in a :py:class:`pylops_mpi.DistributedArray`. To obtain the
 # overall inversion of the entire MPIBlockDiag, you can utilize the ``asarray()``
 # function of the DistributedArray as shown below.
-xinv, istop, niter, r1norm, r2norm, cost = pylops_mpi.cgls(BDiag, y, niter=15, tol=1e-10, show=True)
+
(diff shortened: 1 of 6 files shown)


mrava87 (approved):
@rohanbabbar04 this looks great, I do not have any comment. I merge :D

PR #102, opened 2024-05-26
Fix datatypes for macos test fails

### LINKED ISSUE #101: numpy.float128 missing
Looks like tests are failing (https://github.com/PyLops/pylops-mpi/actions/runs/9237403210/job/25414331862?pr=100) because:

```
/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/site-packages/numpy/__init__.py:333: in __getattr__
    raise AttributeError("module {!r} has no attribute "
E   AttributeError: module 'numpy' has no attribute 'float128'. Did you mean: 'float16'?
```

I see that `numpy-1.26.4` was installed. I tried doing the same in Colab and I do not have any problem, I worry this may be a `macos` thing as these are the actions that failed (others just stopped...)

Anyhow, perhaps we should downcast everything to just float64/complex128 as we mostly do in pylops too? Or we would need to catch the os and based on that choose what to use (but I am not sure it is worth the effort for float128 that is hardly used in real life - in many applications float32 is more than enough 😃

rohanbabbar04:
Closes #101 
- Change `np.float128` to `np.float64` in tests.
- Change `np.complex256` to `np.complex128` in tests.
- Minor change of rtol to 1e-13

### DIFF
--- a/tests/test_derivative.py
+++ b/tests/test_derivative.py
@@ -14,7 +14,7 @@ par1 = {
     "nz": 600,
     "dz": 1.0,
     "edge": False,
-    "dtype": np.float128,
+    "dtype": np.float64,
     "partition": pylops_mpi.Partition.SCATTER
 }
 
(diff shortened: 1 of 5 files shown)


mrava87 (approved):
Looks good, thanks for handling this so fast. Feel free to merge.

PR #104, opened 2024-06-16
minor: added .pyc to gitignore

mrava87:


mrava87:
@rohanbabbar04 I made a new PR and noticed that tests failed (they run locally)... so I went ahead making a dummy PR with just a small change to the .gitignore file to see if the test failures were due to any new code, but I get the same issue here.

It is unclear to me what happens as I don't see to be able to stop the tests and understand where the error is. Do you mind taking a look as well?

rohanbabbar04:
@mrava87, I got it its because of the new version of `numpy=2.0.0` which is released today and has some major changes...

mrava87:
Oh wow, good catch. I knew numpy v2 was coming but didn't realize it was today. 

Ok, let's go ahead with what you have done (force not to install numpy 2) and later we can revise this... we can install numpy 2 locally and try to figure out what is the problem with the Laplacian test.

Feel free to merge this PR

rohanbabbar04 (approved):

PR #109, opened 2024-10-10
Made changes to paper.md

rohanbabbar04:
- Changes in `paper.md`

### DIFF
--- a/joss/paper.bib
+++ b/joss/paper.bib
@@ -1,5 +1,5 @@
 @article{Ravasi:2020,
-    author = {M. Ravasi and I. Vasconcelos},
+    author = {Ravasi, Matteo and Vasconcelos, Ivan},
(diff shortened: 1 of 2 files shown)


mrava87:
Sure, I'll try to reduce it further. How do i find out how many words I have after I push the changes?

rohanbabbar04:
Great!!
I see the word count is now close to 960 words which is good. Are you done with the updates?
I will read the paper when you are done and merge this PR

mrava87:
Perfect, yes I am 😀 feel free to re-add back any sentence that may be worth having but we took out to shorten the paper given that we are now 40 words or so below the limit

rohanbabbar04:
> Perfect, yes I am 😀 feel free to re-add back any sentence that may be worth having but we took out to shorten the paper given that we are now 40 words or so below the limit

Great!!,
Let me merge this

Issue #113, opened 2024-10-23
'2d' distribution


mrava87:
In this Github issue I want to discuss a new form of parallelism that I came across when trying to extend our MDD tutorial, which could be more universal and useful for other problems. 

Currently, we split the problem across one axis (in the operator) and broadcast model and data. This limits somehow how much we can scale in model and data size. A more flexible approach is to split the problem across two axes: one for the operator, and one for the model/data. This way we can easily scale to large model/data with minimal modifications to the inner working of the operator (leveraging the natural parallelism that we exploited already). I am uploading a cartoon to try to help explain this idea in a more visual way.

In order to enable this distribution pattern, a few modifications are however required at the DistributedArray level, namely:

- [X] a new input, called `mask` is added to the `DistributedArray` object, which allows defining subsets of the DistributedArray where the same information is repeated.
- [X] subcommunicators are created in the `__init__` method of `DistributedArray` using the `base_comm.Split` method based on the provided mask such that we can perform reduce/broadcast operations over a subset of ranks
- [X] a new `_allreduce_subcomm` method is added to allow performing reduce operarations on subcommunicators
- [X] operations like `dot` and `norm` now use the `_allreduce_subcomm` as we do not want to add contributions from repeated portions of a `DistributedArray` (as ultimately, the actual information we care about is stored within each subset, replication is only needed to enable '2d' parallelism)
- [X] `MPIBlockDiag` is now also having a a new input called `mask`, which allows one to choose the output of matvec/rmatvec to have the same masking as the input array... 
- [ ] `MPIVStack` would also benefit from this allowing to apply the `allreduce` method for each subcommunicator. We would need to go between DistributedArray objects with Partition.SCATTER and mask to one with Partition.SCATTER  and no mask.

I will soon push a draft PR and we can go through over it and discuss how best to incorporate this 😄 

![Ranks](https://github.com/user-attachments/assets/11c9ab77-ebfa-4a69-a343-a58d571e766f)

mrava87:
See https://github.com/PyLops/pylops-mpi/tree/2ddistr

hongyx11:
Hi Matteo,  have some questions about figures, 
in the upper part, it's an equation? and lhs is split across the rank in vs dimension?
does green, yellow, purple represents different mpi rank?
how about blue and yellow?
Thanks,

mrava87:
Hi Yuxi, 
[...]

PR #114, opened 2024-10-23
2D-distribution

mrava87:
See https://github.com/PyLops/pylops-mpi/issues/113

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest]
-        python-version: ['3.8', '3.9', '3.10', '3.11', '3.12']
+        python-version: ['3.9', '3.10', '3.11', '3.12']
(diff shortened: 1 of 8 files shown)


mrava87:
@rohanbabbar04 this could do with your review and comments 😄

rohanbabbar04:
> @rohanbabbar04 this could do with your review and comments 😄

I would review this by the end of this week :)

rohanbabbar04:
Sorry,
Give me time till this week
I am onto this...

mrava87:
Any news on this @rohanbabbar04

rohanbabbar04:
I think we should add information about the UNSAFE_BROADCAST in the Warning section of [this page](https://pylops.github.io/pylops-mpi/api/generated/pylops_mpi.DistributedArray.html#pylops_mpi.DistributedArray) to inform users about it. What do you think?

mrava87:
I agree, good point! As we already explain the `BROADCAST`, we can explain that to avoid that behavior one has to use `UNSAFE_BROADCAST` at their own risk (with more responsibility comes higher risk 😜

PR #116, opened 2024-10-30
feat: added dottest

mrava87:
This PR adds the `dottest` routine to perform dot-test of pylops-mpi operators.

The rationale behind this addition is that in some cases one may want to implement new operators that are not available in pylops (perhaps as it does not make sense to implement a serial version of such operators), and one needs access to a dot-test to verify their correctness.


mrava87:
@rohanbabbar04 can you take a look at this, would be good not to let PR sit there for too long 😄

rohanbabbar04 (approved):

rohanbabbar04:
> @rohanbabbar04 can you take a look at this, would be good not to let PR sit there for too long 😄

Yes looks good to me 😃

PR #128, opened 2025-03-10
Update build.yml to support python3.13

rohanbabbar04:
- Add support for python3.13


rohanbabbar04:
Maybe in the future to reduce the tests we can drop python3.9.
Do you think we update the docs deployment version to `3.11`?

mrava87:
Great to see it works on py3.13. And yes, I wouldn’t mind stopping support/test for py3.9

PR #130, opened 2025-05-18
Add NCCL support to DistributedArray

tharittk:
I added the support for NCCL collective calls to DistributedArray.py. These calls include `allReduce`, `allGather`, and `bcast` The user passes the `nccl.cuda.nccl.NcclCommunicator` object instead of `MPI.COMM_WORLD` to the `DistributedArray` constructor to use NCCL

### Setup
fork from the main branch and install cupy with nccl to the environment
`$conda env create -f environment-dev.yml`
`$conda activate pylops_mpi`
`$conda install -c conda-forge cupy cudnn cutensor nccl`
`$pip install -e .`

### Change Made

DistributedArray
+ `base_comm` now take either `mpi4py.MPI.Comm` or `NcclCommunicator`
+ `sub_comm` assignment is carried through the `subcomm_split()`  `cupy nccl` doesn't provide split API as its C++ counterpart so this function implements the split manually.
+ add `_allgather` - MPI provides `allgather` (returns value) and `Allgather` (requires recv buffer) while `cupy nccl` only provides `allGather` that requires recv buffer. This function provides unified API for both cases of `base_comm`. Now the the call is `self._allgather()` instead of `self.basecomm.allgather` - which only worked with MPI

+ add NCCL support to `_allreduce` and `_allreduce_subcomm`
+ `bcast` in `__setitem__` can also use `NcclCommunicator` counterpart
+ Explicitly use `MPI.COMM_WORLD.Get_size()` and `MPI.COMM_WORLD.Get_rank()` in `rank()` and `size()` getter instead of `base_comm`. This is because NCCL recommends to launch with one MPI process per GPU -- violates this may cause deadlock. This means in either MPI and NCCL, the global rank is managed through MPI. Plus, the `cupy nccl` does not provide rank getter API.

+ `dot()` and `_compute_vector_norm()` now use `ncp = get_module(self.engine)` to reflect that computation may now be done in GPU memory `cupy array`

test_distributedarray_nccl.py
+ Almost exact copy of `test_distributedarray.py` but now test under `base_comm` is `cupy.cuda.nccl.NcclCommunicator`
+ NCCL does not support the `dtype=complex` so some test cases involve complex number are removed
+ in order to call `assert_allclose`, the `.get()` is called because under NCCL, the array lives inside GPU memory (and did not implicitly copy it CPU)

### Testing
the test is done in two scenarios in NCSA Delta A40x40 
+ 1 node with multi-gpu (2 and 3 GPUs) with following commands
`srun -N 1 -n 3  --gpus-per-node=3 pytest --with-mpi` for 3 GPUs
or in local workstation with 3 GPUs
`mpirun -n 3 pytest --with-mpi`
+ 2 nodes with multi-gpu (2 and GPUs per node)
`srun -N 2 -n 6  --gpus-per-node=3 pytest --with-mpi`


pylops_mpi/DistributedArray.py
```diff
@@ -1,12 +1,39 @@
-import numpy as np
-from typing import Optional, Union, Tuple, List
-from numbers import Integral
-from mpi4py import MPI
 from enum import Enum
+from numbers import Integral
+from typing import List, Optional, Tuple, Union
 
+import cupy as cp
+import cupy.cuda.nccl as nccl
```
mrava87, line 6: We cannot import these directly as this will lead to an import error when we import pylops_mpi in an environment without `cupy` and `nccl`. 

I suggest to create a `deps` and `backend` files in the `utils` module similar to those in PyLops and use them to create protected imports

pylops_mpi/DistributedArray.py
```diff
@@ -1,12 +1,39 @@
...
+import numpy as np
+from mpi4py import MPI
 from pylops.utils import DTypeLike, NDArray
 from pylops.utils._internal import _value_or_sized_to_tuple
-from pylops.utils.backend import get_module, get_array_module, get_module_name
+from pylops.utils.backend import get_array_module, get_module, get_module_name
+
+cupy_to_nccl_dtype = {
```
mrava87, line 13: Similar to the above, I would suggest adding maybe a `nccl` file in `utils` where we add all NCCL related functions (or dicts like this one)...

pylops_mpi/DistributedArray.py
```diff
@@ -1,12 +1,39 @@
...
+    "int32": nccl.NCCL_INT32,
+    "int64": nccl.NCCL_INT64,
+    "uint8": nccl.NCCL_UINT8,
+    "int8": nccl.NCCL_INT8,
+    "uint32": nccl.NCCL_UINT32,
+    "uint64": nccl.NCCL_UINT64,
+}
+
+
+# MPI.Op is not hashable and cannot convert to string to use dictionary map
```
mrava87, line 25: And this one.... Also move the `# ...` comment as a docstring of the function; also try to always add some description of the parameters and returns of a function in the doctoring

pylops_mpi/DistributedArray.py
```diff
@@ -49,14 +76,53 @@ def local_split(global_shape: Tuple, base_comm: MPI.Comm,
         local_shape = global_shape
     # Split the array
     else:
+        # TODO: Both NCCL and MPI have to use MPI.COMM_WORLD for shape splitting
+        # So I decide to make it explicit. This will make base_comm unused.
```
mrava87, line 80: Try not to make comments in first person. You can say `it was decided`

Apart from this stylistic comment, I am not sure I understand this. Why can we not keep using `base_comm`? I guess because if you want to use the NCCL backend this will be `cupy.cuda.nccl.NcclCommunicator` but we want to use the MPI one in any case?

In general, I would expect to always use MPI for passing around small scalars that define shapes (you would unlikely move those onto a GPU and use NCCL anyways), agree?

pylops_mpi/DistributedArray.py
```diff
@@ -49,14 +76,53 @@ def local_split(global_shape: Tuple, base_comm: MPI.Comm,
...
+    ----------
+    mask : :obj:`list`
+        Mask defining subsets of ranks to consider when performing 'global'
+        operations on the distributed array such as dot product or norm.
+
+    base_comm : :obj:`mpi4py.MPI.Comm`, optional
+        A Communicator over which array is distributed
+        Defaults to ``mpi4py.MPI.COMM_WORLD``.
+
+    Returns:
```
mrava87, line 105: Returns
--------

pylops_mpi/DistributedArray.py
```diff
@@ -49,14 +76,53 @@ def local_split(global_shape: Tuple, base_comm: MPI.Comm,
...
+        sub_rank = sub_comm.Get_rank()
+        sub_size = sub_comm.Get_size()
+
+        if sub_rank == 0:
+            nccl_id_bytes = nccl.get_unique_id()
+        else:
+            nccl_id_bytes = None
+        nccl_id_bytes = sub_comm.bcast(nccl_id_bytes, root=0)
+        nccl_subcomm = nccl.NcclCommunicator(sub_size, nccl_id_bytes, sub_rank)
+        return nccl_subcomm
```
mrava87, line 121: I suggest you store the output into a single `subcomm` in both branches and then return `subcomm`

pylops_mpi/DistributedArray.py
```diff
@@ -49,14 +76,53 @@ def local_split(global_shape: Tuple, base_comm: MPI.Comm,
...
+            local_shape[axis] = global_shape[axis] // MPI.COMM_WORLD.Get_size() + 1
         else:
-            local_shape[axis] = global_shape[axis] // base_comm.Get_size()
+            local_shape[axis] = global_shape[axis] // MPI.COMM_WORLD.Get_size()
     return tuple(local_shape)
 
 
+def subcomm_split(mask, base_comm: MPI.Comm):
+    """To create the subcommunication based on mask.
+    CuPy's NCCL does not support ncclCommSplit as it C++ API does.
```
mrava87, line 91:
```
To create the subcommunication based on mask.
CuPy's NCCL does not support ncclCommSplit as it C++ API does.
```

into:

```
Create new communicators based on mask

This method creates new NCCL communicators based on ``mask``. Contrary to MPI, NCCL does not provide support for splitting of a communicator in multiple subcommunicators; this is therefore handled explicitly by this method.
```

pylops_mpi/DistributedArray.py
```diff
@@ -273,7 +358,12 @@ def local_shapes(self):
         -------
         local_shapes : :obj:`list`
         """
-        return self.base_comm.allgather(self.local_shape)
+        if self.base_comm is MPI.COMM_WORLD:
```
mrava87, line 361: Above you have `if isinstance(self.base_comm, nccl.NcclCommunicator):`, I think it would be good to always have the same pattern, and I personally prefer:

  if self.base_comm is MPI.COMM_WORLD:
```
if isinstance(self.base_comm, MPI.Comm):
    # MPI
else:
    # NCCL
```
to be checked...

pylops_mpi/DistributedArray.py
```diff
@@ -387,22 +509,83 @@ def _check_mask(self, dist_array):
...
+            return recv_buf
+        else:
+            if recv_buf is None:
+                return self.sub_comm.allreduce(send_buf, op)
+            # For MIN and MAX which require recv_buf
+            self.sub_comm.Allreduce(send_buf, recv_buf, op)
+            return recv_buf
+
+    def _allgather(self, send_buf, recv_buf=None):
+        """Allgather oepration
```
mrava87, line 564: oepration -> operation

pylops_mpi/DistributedArray.py
```diff
@@ -273,7 +358,12 @@ def local_shapes(self):
         -------
         local_shapes : :obj:`list`
         """
-        return self.base_comm.allgather(self.local_shape)
+        if self.base_comm is MPI.COMM_WORLD:
+            return self._allgather(self.local_shape)
+        else:
+            size = len(self.local_shape)
```
mrava87, line 364: explain a bit this code pattern, it is not obvious

mrava87 (changes requested):
Hi @tharittk,
Great start!

I have started reviewing your PR and I am providing here a number of general comments, I will then make more specific ones in the code itself:

- one of the successfull decisions that we made in PyLops since the beginning was to limit the number of mandatory dependencies to the minimum (basically numpy and scipy) and allow a lot of optional dependencies which a user can choose to install or not, whilst still being able to use the library. Similar for `pylops-mpi` we extended the mandatory dependencies to `mpi4py` for obvious reasons but no more than that. In our current implementation of `DistributedArray` we support a double backend via the engine keyword (numpy and cupy), however we never import `cupy` - this is possible because we use the methods `get_module, get_array_module, get_module_name` in the module `pylops.utils.backend` - have a look at them to understand what they do. I will also provide more detailed comments in the code.
- Whilst we now want to provide a NCCL backend using `cupy`, I still believe that we should try not to make `cupy` a mandatory dependency, also because there are various flavors of cupy installations based on the CUDA drivers so we cannot just add one to the dependencies in the `pyproject.toml` file. 
- Based on the above, I suggesting adding the following to the `pyproject.toml` file:
```
[project.optional-dependencies]
cupy11 = [
    "cupy-cuda11x",
]
cupy12 = [
    "cupy-cuda12x",
]
```
to be tested. We should also add `nccl` but I see that there are a lot in PyPI so we need to check which one is the correct one. 
- I very much like your detailed description of the setup in the PR. I suggest you have a go at updating `README.md` and `docs/source/installation.rst`. Here you could do like in the current files, add new targets to `MakeFile` with your installation instructions and refer to them.
- Same for the tests, you can add some text in the  `docs/source/installation.rst` file (both the `mpirun` and `srun` as these are valid for any SLURM cluster and not really NCSA Delta specific)

pylops_mpi/DistributedArray.py
tharittk: Yes, MPI is expected to do that for small scalars. `cupy.cuda.nccl.NcclCommunicator`  has the function `rank_id()`  like `Get_rank()` in MPI as well. I would say the `global_shape` is not moved to GPU, so we might as well explicitly use MPI. 

The comment may be misleading as it is not "have to".

tharittk: Actually, the `local_split()` does not trigger any communication calls. It uses those `rank` and `size` for calculation. It maybe better style to use `base_comm` as you suggested.

pylops_mpi/DistributedArray.py
tharittk: That sounds good. I will do that fix.

pylops_mpi/DistributedArray.py
tharittk: Ah yes, I think I saw the discussion somewhere in PyLops about the preference for single-return style. I will make that change.

tests/test_distributedarray_nccl.py
```diff
@@ -0,0 +1,438 @@
+"""Test the DistributedArray class
```
mrava87, line 1: Another general comment about tests: these look great and I think they do exactly what I would envision NCCL tests to do - map nearly 1-to-1 the equivalent MPI test so checking against serial code results 😄.

However, you would see if I trigger our CI (it will usually trigger automatically but since you are first time contributor I have to allow it for once) that this file would make it break as it imports cupy which is not in our `requirements-dev.txt` file (https://github.com/PyLops/pylops-mpi/blob/f1bc646f5a0d3d8790159a00d08fc897f0d5b87a/.github/workflows/build.yml#L42). For now my suggestion is to make a new folder called `test_nccl` and populate with files with NCCL tests. We can then modify the Makefile adding:

```
tests_nccl:
	mpirun -n $(NUM_PROCESSES) pytest tests_nccl/ --with-mpi
```
(note that we have `mpiexec` in all our documentation but you use `mpirun`, any specific reason?)

and here https://github.com/PyLops/pylops-mpi/blob/f1bc646f5a0d3d8790159a00d08fc897f0d5b87a/.github/workflows/build.yml#L46 with:

```
 run: mpiexec -n ${{ matrix.rank }} pytest tests/ --with-mpi
```

tests/test_distributedarray_nccl.py
```diff
@@ -0,0 +1,438 @@
...
+
+import numpy as np
+from mpi4py import MPI
+import pytest
+from numpy.testing import assert_allclose
+
+from pylops_mpi import DistributedArray, Partition
+from pylops_mpi.DistributedArray import local_split
+
+import cupy as cp
```
mrava87, line 16: This is fine to import this way as we know we will never run this test without `cupy`... in future we may however want to do something like this https://github.com/PyLops/pylops/pull/660 and we will eventually create a Github-action for the NCCL backend with the help of @hongyx11  like we are about to do for PyLops

But let's not do this until the NCCL API is a bit more evolved...

tests/test_distributedarray_nccl.py
```diff
@@ -0,0 +1,438 @@
...
+np.random.seed(42)
+
+comm = MPI.COMM_WORLD
+rank = comm.Get_rank()
+size = comm.Get_size()
+host = MPI.Get_processor_name()
+
+
+# Use MPI to set-up NCCL Communication
+def initialize_nccl_comm():
```
mrava87, line 29: Just a question as it may be the case, but is it really needed to get these slurm variables or you could always discover it differently (I guess the other two cases? The reason I am asking is because you say tests can be run with `mpirun -n 3 pytest --with-mpi` as well and there the variable will be `None`.

tests/test_distributedarray_nccl.py
```diff
@@ -0,0 +1,438 @@
...
+        nccl_id = None
+    nccl_id = comm.bcast(nccl_id, root=0)
+
+    nccl_comm = nccl.NcclCommunicator(size, nccl_id, rank)
+    return nccl_comm
+
+
+nccl_comm = initialize_nccl_comm()
+
+# NCCL Does not support complex type
```
mrava87, line 50: Interesting, I wasn't aware of this... @hongyx11 any idea if this is in the roadmap of NCCL?

pylops_mpi/DistributedArray.py
mrava87: Mmh but you use `Get_rank` and `Get_size` in the function, don't you? So perhaps we really need to use `MPI.COMM_WORLD` because of the different function signatures between MPI and NCCL?

tharittk: Yes, we can do that in the `backend.py`. I think it will simply be `if-else` of MPI and NCCL and call the appropriate function for each.

I did some search and actually the C++ NCCL API does not have this get_rank() function. Looks like the rank is forced to be accessed through MPI. Maybe either choice will not make any functional difference.

tests/test_distributedarray_nccl.py
tharittk: Oh that's my bad. I was not careful regarding `mpirun` and `mpiexec`. I should have called `mpiexec` to be consistent.

tests/test_distributedarray_nccl.py
tharittk: I think I can test that by removing the `os.environ.get("SLURM_LOCALID")` and run on the Delta in multi-node w/ multi-gpu environment. But yeah, the initial attempt was like to exhaustively cover where the variable may exist (and thus 3 checks)

pylops_mpi/utils/deps.py
```diff
@@ -0,0 +1,41 @@
...
+
+import os
+from importlib import import_module, util
+from typing import Optional
+
+
+# error message at import of available package
+def nccl_import(message: Optional[str] = None) -> str:
+    nccl_test = (
+        # TODO: setting of OS env should go to READme somewhere
```
mrava87, line 13: This is explained here in the documentation of PyLops: https://pylops.readthedocs.io/en/stable/gpu.html 

We could add a similar page to that of PyLops-MPI and in the README simply say we refer to the documentation for details on how to use NCCL.

hongyx11:
- Good news, I was able to run the tests on my local environment. I think it's quite clean and clear.
- @mrava87 complex support is important to PyLops right? we might find other way to do it.

pylops_mpi/utils/__init__.py
```diff
@@ -1,2 +1,6 @@
 # isort: skip_file
-from .dottest import *
+
+# currently dottest create circular dependency with DistributedArray.py
```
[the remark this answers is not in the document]
tharittk: Yes, because now the `DistributedArray.py` calls `from pylops_mpi.utils.backend import cupy_to_nccl_dtype, mpi_op_to_nccl` and the `dottest.py` also then `import DistributedArray` - and that recreats circular dependency. The program will just throw the error.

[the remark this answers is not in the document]
mrava87: Alright, I see. Then it’s ok to remove this, simply means we need a bit longer signature to import dottest, but I don’t see that as a problem 😀

I checked why we did not have a similar problem in pylops and that’s because we don’t have a object container like DistributedArray and so we just import numpy in dottest.py

mrava87:
> * Good news, I was able to run the tests on my local environment. I think it's quite clean and clear.
> * @mrava87 complex support is important to PyLops right? we might find other way to do it.

@hongyx11 for sure there are interesting scientific problems where the data and/or model is complex valued. For now I think we can leave this out but I think it would make it a nice PR for @tharittk later on in the GSoC project to handle this: probably the easiest will be to pack/unpack c on complex arrays into real/imag concatenations… I think we may be able to do this with a decorator that we add to all methods that do communications 😀

pylops_mpi/utils/backend.py
```diff
@@ -0,0 +1,282 @@
...
+            (0, send_shape[i] - local_array.shape[i])
+            for i in range(len(send_shape))
+        ]
+
+        send_buf = cp.pad(
+            local_array, pad_size, mode="constant", constant_values=0
+        )
+
+        # NCCL recommends to use one MPI Process per GPU
+        ndev = MPI.COMM_WORLD.Get_size()
```
mrava87, line 264: can this not be passed to the function, instead of refinding it inside... in `DistributedArray` which is where so far we call this method, it is already in `self.size`

tests_nccl/test_distributedarray_nccl.py
```diff
@@ -0,0 +1,412 @@
...
+import numpy as np
+from mpi4py import MPI
+import pytest
+from numpy.testing import assert_allclose
+
+from pylops_mpi import DistributedArray, Partition
+from pylops_mpi.DistributedArray import local_split
+from pylops_mpi.utils.backend import initialize_nccl_comm, get_cupy
+
+cp = get_cupy()
```
mrava87, line 17: no need to do this in tests, tests are developer facing and we know that if we run this test we need cupy... it is only for user facing code that we want to allow one to use pylops-mpi with and without cupy/nccl

I think the same may apply to `initialize_nccl_comm` but since that is a bit more complex method that does quite a few things and may be also useful in scripts, I will keep where it is 😄

pylops_mpi/utils/backend.py
```diff
@@ -0,0 +1,282 @@
...
+        ]
+
+        for i in range(ndev):
+            slicing = tuple(slice(0, end) for end in local_shapes[i])
+            chunks[i] = chunks[i].reshape(send_shape)[slicing]
+        final_array = cp.concatenate([chunks[i] for i in range(ndev)], axis=axis)
+        return final_array
+
+
+def get_cupy() -> ModuleType:
```
mrava87, line 281: See comment in the test file, basically this is not needed...

pylops_mpi/utils/backend.py
```diff
@@ -0,0 +1,282 @@
...
+            local_array, pad_size, mode="constant", constant_values=0
+        )
+
+        # NCCL recommends to use one MPI Process per GPU
+        ndev = MPI.COMM_WORLD.Get_size()
+        recv_buf = cp.zeros(ndev * send_buf.size, dtype=send_buf.dtype)
+        # self._allgather(send_buf, recv_buf)
+        nccl_allgather(nccl_comm, send_buf, recv_buf)
+
+        chunk_size = cp.prod(cp.asarray(send_shape))
```
mrava87, line 269: no need to move this to the gpu, just use numpy... on the gpu we want to do the key operations with large arrays, not small stuff dealing with indices etc.

pylops_mpi/utils/backend.py
```diff
@@ -0,0 +1,282 @@
...
+        send_buf = cp.pad(
+            local_array, pad_size, mode="constant", constant_values=0
+        )
+
+        # NCCL recommends to use one MPI Process per GPU
+        ndev = MPI.COMM_WORLD.Get_size()
+        recv_buf = cp.zeros(ndev * send_buf.size, dtype=send_buf.dtype)
+        # self._allgather(send_buf, recv_buf)
+        nccl_allgather(nccl_comm, send_buf, recv_buf)
+
```
mrava87, line 268: Add a comment, as it me a while to understand what you were doing, something like:

```
# extract individual arrays, remove padding, and combine them back into a single global array
```

pylops_mpi/utils/backend.py
```diff
@@ -0,0 +1,282 @@
...
+        chunk_size = cp.prod(cp.asarray(send_shape))
+        chunks = [
+            recv_buf[i * chunk_size:(i + 1) * chunk_size] for i in range(ndev)
+        ]
+
+        for i in range(ndev):
+            slicing = tuple(slice(0, end) for end in local_shapes[i])
```
mrava87, line 275: do you need slicing here, as you know you will always have to do:

```
chunks[i] = chunks[i].reshape(send_shape)[:local_shapes[i]]
```

pylops_mpi/utils/backend.py
```diff
@@ -0,0 +1,282 @@
...
+            chunks[i] = chunks[i].reshape(send_shape)[slicing]
```
mrava87, line 276: why is this needed `[chunks[i] for i in range(ndev)]`, chunks is already a list?

tests_nccl/test_distributedarray_nccl.py
```diff
@@ -0,0 +1,412 @@
...
+
+pytestmark = pytest.mark.skipif(cp is None, reason="CuPy not available")
```
mrava87, line 19: Seems unused?

pylops_mpi/DistributedArray.py
```diff
@@ -49,14 +50,45 @@ def local_split(global_shape: Tuple, base_comm: MPI.Comm,
...
+        Mask defining subsets of ranks to consider when performing 'global'
+        operations on the distributed array such as dot product or norm.
+
+    base_comm : :obj:`mpi4py.MPI.Comm`, optional
+        A Communicator over which array is distributed
+        Defaults to ``mpi4py.MPI.COMM_WORLD``.
+
+    Returns:
+    -------
+        Union[mpi4py.MPI.Comm, cupy.cuda.nccl.NcclCommunicator]]: a subcommunicator according to mask
```
mrava87, line 81: see comment in other places about conforming to the doctoring style of pylops/pylops-mpi... this should be

```
sub_comm : :obj:`mpi4py.MPI.Comm` or :obj:`cupy.cuda.nccl.NcclCommunicator`
        Subcommunicator according to mask
```

pylops_mpi/DistributedArray.py
```diff
@@ -96,8 +128,9 @@ class DistributedArray:
         Type of elements in input array. Defaults to ``numpy.float64``.
     """
 
+    # TODO: Type Annotation for base_comm without NCCL import
     def __init__(self, global_shape: Union[Tuple, Integral],
-                 base_comm: Optional[MPI.Comm] = MPI.COMM_WORLD,
+                 base_comm=MPI.COMM_WORLD,
```
mrava87, line 133: look at this example for how to add type annotation of a package that may not be present: https://github.com/PyLops/pylops/blob/44bae08e43c34516b71ac08e4e445e5835ddcc47/pylops/waveeqprocessing/twoway.py#L14

mrava87:
@tharittk I also triggered the CI and as you can see the actions fail. But I think I have identified the problem 😄 

The way `util.find_spec` works is that it tries to go all the way to nccl assuming cupy is present, but since it is not it fails
```
util.find_spec("cupy.cuda.nccl")
```

If instead you do
```
 util.find_spec("cupy") is not None and util.find_spec("cupy.cuda.nccl") is not None
```

it will work as it will stop already at the first check and if cupy isn't present will return False... try to change it and we will see if this is the only problem or if more will arise (from now on the CI will be automatically triggered at any push you do)

tests_nccl/test_distributedarray_nccl.py
tharittk: Possibly, I put it there in case the pytest accidentally triggers this folder. But it maybe unnecessary if the Makefile specifies to run only tests/.

pylops_mpi/utils/backend.py
```diff
@@ -0,0 +1,282 @@
...
+        """
+        comm = MPI.COMM_WORLD
+        rank = comm.Get_rank()
+        size = comm.Get_size()
+        local_rank = int(
+            os.environ.get("SLURM_LOCALID")
+            or os.environ.get("OMPI_COMM_WORLD_LOCAL_RANK")
+            or rank % cp.cuda.runtime.getDeviceCount()
+        )
+        device_id = local_rank
```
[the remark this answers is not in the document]
tharittk: I guess you are right. I was probably too sensitive to the naming back then i.e., `local_rank` is more like MPI lingo while `device_id` is GPU lingo and I was trying to separate that.

pylops_mpi/utils/backend.py
```diff
@@ -0,0 +1,282 @@
...
+        cp.cuda.Device(device_id).use()
```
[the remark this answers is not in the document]
tharittk: oh this is something new to me actually. Let me take a look.

pylops_mpi/utils/backend.py
tharittk: I think we need it. It was a bug I found previously. 
For multi-dimensional data, NCCL returns the flat array. If the padded dimension is along axis != 0, the zeros from padding may appear somewhere inside the flat array (and not at the tail - as it may be in case axis=0). 

So truncating it with `[:local_shapes[i]]` may not be correct.  Slicing after reshape will get rid of padded zeros in the correct dimension

pylops_mpi/utils/backend.py
tharittk: Yes chunks is already a list. But I think if we need to do slicing (as comment above), we need this for-loop and modify the `chunks[I]` in-place.

mrava87: Oh ok, I see now what you are doing, makes sense as you may have pad on a different axis than 0

pylops_mpi/utils/backend.py
mrava87: Oh I see, makes senss

pylops_mpi/utils/backend.py
```diff
@@ -0,0 +1,282 @@
...
+        "float64": nccl.NCCL_FLOAT64,
+        "int32": nccl.NCCL_INT32,
+        "int64": nccl.NCCL_INT64,
+        "uint8": nccl.NCCL_UINT8,
+        "int8": nccl.NCCL_INT8,
+        "uint32": nccl.NCCL_UINT32,
+        "uint64": nccl.NCCL_UINT64,
+    })
+
+    class NcclOp(IntEnum):
```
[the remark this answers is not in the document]
mrava87: Oh I see, I didn't try and assumed will be like for numpy functions... let's leave it as is then 😄

tests_nccl/test_distributedarray_nccl.py
tharittk: I see. That makes total sense.

pylops_mpi/utils/backend.py
tharittk: Oh yes. Absolutely we don't need that.

pylops_mpi/DistributedArray.py
```diff
@@ -1,12 +1,13 @@
...
+from numbers import Integral
+from typing import List, Optional, Tuple, Union
 
+import numpy as np
+from mpi4py import MPI
 from pylops.utils import DTypeLike, NDArray
 from pylops.utils._internal import _value_or_sized_to_tuple
-from pylops.utils.backend import get_module, get_array_module, get_module_name
+from pylops.utils.backend import get_array_module, get_module, get_module_name
+from pylops_mpi.utils.backend import nccl_split, nccl_allgather, nccl_allreduce, nccl_bcast, nccl_asarray
```
rohanbabbar04, line 10: This will currently break if I do not have `nccl` and `cupy` on my system.

```suggestion
try:
    from pylops_mpi.utils.backend import (
        nccl_split,
        nccl_allgather,
        nccl_allreduce,
        nccl_bcast,
        nccl_asarray,
    )
except ImportError:
    pass
```

pylops_mpi/utils/backend.py
```diff
@@ -0,0 +1,282 @@
+__all__ = [
+    "initialize_nccl_comm",
+    "get_cupy",
+    "nccl_split",
+    "nccl_allgather",
+    "nccl_allreduce",
+    "nccl_bcast",
+    "nccl_asarray"
+]
```
rohanbabbar04, line 9: This will not be recognized outside the `if` block, and will lead to errors.

rohanbabbar04 (changes requested):
**Overall, it looks great!**

@tharittk, some general comments for you to address, but it's a great start.

@mrava87, regarding the case of avoiding `MPI.COMM_WORLD`, I think it might be better to introduce `base_comm_nccl`, and we consistently use `self.base_comm` everywhere instead of hardcoding `MPI.COMM_WORLD`.

We can do something like this for now, 
```python
# Check and override base_comm with MPI.COMM_WORLD for now
if base_comm_nccl is not None and isinstance(base_comm_nccl, NCCLCommunicator):
    self._base_comm = MPI.COMM_WORLD
    self._base_comm_nccl = base_comm_nccl
```

pylops_mpi/DistributedArray.py
mrava87: The reason it fails is likely due to the comments I put somewhere about having to change https://github.com/tharittk/pylops-mpi/blob/f1238cb0c26a90b672e2db38effb701dea1accee/pylops_mpi/utils/deps.py#L14 with a first condition checking if cupy is present.. I believe with that we should be good and can use the same code pattern to import those methods as for example here: https://github.com/PyLops/pylops/blob/44bae08e43c34516b71ac08e4e445e5835ddcc47/pylops/waveeqprocessing/twoway.py#L12

pylops_mpi/utils/backend.py
mrava87: I don’t think so, we do the exact same in pylops (https://github.com/PyLops/pylops/blob/dev/pylops/utils/backend.py) and everything works even when cupy is not installed… we are probably not just following the same code pattern yet here 😉

mrava87:
> **Overall, it looks great!**
> 
> @tharittk, some general comments for you to address, but it's a great start.
> 
> @mrava87, regarding the case of avoiding `MPI.COMM_WORLD`, I think it might be better to introduce `base_comm_nccl`, and we consistently use `self.base_comm` everywhere instead of hardcoding `MPI.COMM_WORLD`.
> 
> We can do something like this for now,
> 
> ```python
> # Check and override base_comm with MPI.COMM_WORLD for now
> if base_comm_nccl is not None and isinstance(base_comm_nccl, NCCLCommunicator):
>     self._base_comm = MPI.COMM_WORLD
>     self._base_comm_nccl = base_comm_nccl
> ```

Thanks @rohanbabbar04 for taking the time to review this first PR of @tharittk 

I agree with your judgment on the MPI.COMM_WORLD usage: let’s try to go for the option of having two base_comm inputs and use NCCL if both are set… @tharittk try to have a go at it and we can discuss all together to make sure it’s done in the best way to ensure backward compatibility also in edge scenarios 😀

tests/test_distributedarray_nccl.py
tharittk: Without `os.environ.get("SLURM_LOCALID")`, the test still passes on Delta in multi-node w/ multi-gpu. So I think we can remove it. The `os.environ.get("OMPI_COMM_WORLD_LOCAL_RANK")` will work if it uses OpenMPI. But the `rank % cp.cuda.runtime.getDeviceCount()` will work in any case.

pylops_mpi/DistributedArray.py
rohanbabbar04: Hmm, yup I forgot we already have `deps.nccl_import`, just updated the commit suggestion. It should now work fine after changing https://github.com/tharittk/pylops-mpi/blob/f1238cb0c26a90b672e2db38effb701dea1accee/pylops_mpi/utils/deps.py#L14 and applying the above suggestion change.

pylops_mpi/utils/backend.py
tharittk: I see. That makes sense.

pylops_mpi/utils/backend.py
```diff
@@ -0,0 +1,282 @@
...
+
+        Returns
+        -------
+        final_array : :obj:`cupy.ndarray`
+            Global array gathered from all GPUs and concatenated along `axis`.
+        """
+        sizes_each_dim = list(zip(*local_shapes))
+        # NCCL allGather requires the send_buf to have the same size for every device
+        send_shape = tuple(map(max, sizes_each_dim))
+        pad_size = [
```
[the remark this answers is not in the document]
tharittk: Yes this works. Only the `for l_shape, s_shape` has to be switched to `for s_shape, l_shape`

pylops_mpi/utils/backend.py
rohanbabbar04: Maybe we can create a common function

```python
# Keep MPI.COMM_WORLD as default
def allgather(comm=MPI.COMM_WORLD, send_buf=None, recv_buf=None):
   if isinstance(comm, MPI.COMM):
       .....
   elif deps.nccl_enabled:
       # Add code for NCCL
```

We can do the same for other functions as well. In this way, we do not need to update the _all_gather, _all_reduce, etc in DistributedArray class and in the future can be used inside other functions as well

mrava87: This was the initial idea I suggested 😀 I think it is doable as long as we find a good way at that level to again be protected against imports of cupy/nccl… so basically we need to keep all we have so far but then add the allgather you show in the backend file

pylops_mpi/DistributedArray.py
tharittk: This one is solved once we introduce both `base_comm` and `base_comm_nccl`. In this, we take in `MPI.Comm` so we don't have to account for the scenario in which NCCL calls may be needed.

pylops_mpi/utils/backend.py
tharittk: Currently the function signature is `nccl_asarray(nccl_comm, local_array, local_shapes, axis)` and `ndev` comes from `len(local_shapes)`

hongyx11:
The code looks great! @mrava87 @tharittk

rohanbabbar04 (changes requested):
https://github.com/tharittk/pylops-mpi/blob/3dc41fe740b2a608fa909692bab8a8d5987968fa/pylops_mpi/utils/dottest.py#L7

You can replace this line with, should work and then you can uncomment the dottest

```python
from pylops_mpi import DistributedArray
```

PR #134, opened 2025-05-29
Fix dottest dependency and use base_comm instead of MPI.COMM_WORLD

rohanbabbar04:
- Replace `MPI.COMM_WORLD` with `self.base_comm`
- Fix `dottest` in `__init__.py`
- Fix flake8 lint

### DIFF
--- a/pylops_mpi/DistributedArray.py
+++ b/pylops_mpi/DistributedArray.py
@@ -256,7 +256,7 @@ class DistributedArray:
 
         Returns
         -------
-        engine : :obj:`list`
+        mask : :obj:`list`
         """
         return self._mask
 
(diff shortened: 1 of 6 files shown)


mrava87 (approved):
Hi @rohanbabbar04, thanks for these updates. 

For the minor ones, I am totally fine; always good to fix things here and there that are incorrect 😄 

Regarding `MPI.COMM_WORLD` vs `self.base_comm` I believe your changes are good in the light of the fact that we have now to comm inputs to DistributedArray. But I just want to check with @tharittk that he also agrees before merging this

tharittk:
@mrava87 I agree with the changes. Actually, those changes were suggested by @rohanbabbar04 in #130, but I was unaware of them (they slipped my mind), so I missed fixing those places.

rohanbabbar04:
Great, thanks @mrava87 and @tharittk for your reviews! I'll go ahead and merge this.

PR #137, opened 2025-06-03
add NCCL support to  add_ghost_cells and operators in /basicoperators

tharittk:
I added the support for NCCL to the MPIVStack. Along the way, I discover some interesting issue and it may be worth discussing so I initiate this drafted PR

Change Made
DistributedArray.py

 - MPIVStack adjoint operation takes decorator `@reshape` which calls `add_ghost_cells()` in this file. `add_ghost_cells()` has to be modified to support NCCL. There two points I want to point out.
 1. the call to `self._allgather(cell_fronts)`: this `cell_fronts` are the metadata and is small in size (list of ints, size = total ranks). Under the current implementation, I will dispatch to call NCCL if NCCL is enabled. Should we enforce it to always use MPI instead ?
 2. the call to `self.base_comm.send`: this sends the ghost cells to peer and the sent array can be large. Under the current code, it will always use MPI. If we want to have it use NCCL, we will need to implement point-to-point NCCL. Luckily, CuPy NCCL supports this. I think there will be just adding more call in `_nccl.py`

VStack.py
- the operator now takes one option argument `base_comm_nccl` - just like what we did in `DistrbiutedArray`. I did not change the `MPILinearOperator` interface to take this argument though. But I don't have a strong opinion on either case.
- the output `y` is from `Op @ x` and `Op.H @ x` now initialized to have the same base_comm_nccl as `x` i.e., if `x` lives in NCCL, `y` should lives and communicate with NCCL too.

test_stack_nccl.py
- this closely follows its counterpart MPI test_stack.py but tests explicitly in NCCL environment
- [x] script for testing `HStack` operator

### DIFF
--- a/pylops_mpi/DistributedArray.py
+++ b/pylops_mpi/DistributedArray.py
@@ -748,7 +748,13 @@ class DistributedArray:
         """
         ghosted_array = self.local_array.copy()
         if cells_front is not None:
-            total_cells_front = self._allgather(cells_front) + [0]
+            # TODO: these are metadata (small size). Under current API, it will
+            # call nccl allgather, should we force it to always use MPI?
+            cells_fronts = self._allgather(cells_front)
+            if deps.nccl_enabled and getattr(self, "base_comm_nccl"):
+                total_cells_front = cells_fronts.tolist() + [0]
+            else:
+                total_cells_front = cells_fronts + [0]
             # Read cells_front which needs to be sent to rank + 1(cells_front for rank + 1)
             cells_front = total_cells_front[self.rank + 1]
             if self.rank != 0:
@@ -761,10 +767,16 @@ class DistributedArray:
                                      f"{self.local_shape[self.axis]} < {cells_front}; "
                                      f"to achieve this use NUM_PROCESSES <= "
                                      f"{max(1, self.global_shape[self.axis] // cells_front)}")
+                # TODO: this array maybe large. Currently it will always use MPI.
+                # Should we enable NCCL point-point here ?
                 self.base_comm.send(np.take(self.local_array, np.arange(-cells_front, 0), axis=self.axis),
                                     dest=self.rank + 1, tag=1)
         if cells_back is not None:
-            total_cells_back = self._allgather(cells_back) + [0]
+            cells_backs = self._allgather(cells_back)
+            if deps.nccl_enabled and getattr(self, "base_comm_nccl"):
+                total_cells_back = cells_backs.tolist() + [0]
+            else:
+                total_cells_back = cells_backs + [0]
             # Read cells_back which needs to be sent to rank - 1(cells_back for rank - 1)
             cells_back = total_cells_back[self.rank - 1]
             if self.rank != 0:
--- a/pylops_mpi/DistributedArray.py
+++ b/pylops_mpi/DistributedArray.py
@@ -14,7 +14,7 @@ cupy_message = pylops_deps.cupy_import("the DistributedArray module")
 nccl_message = deps.nccl_import("the DistributedArray module")
 
 if nccl_message is None and cupy_message is None:
-    from pylops_mpi.utils._nccl import nccl_allgather, nccl_allreduce, nccl_asarray, nccl_bcast, nccl_split
+    from pylops_mpi.utils._nccl import nccl_allgather, nccl_allreduce, nccl_asarray, nccl_bcast, nccl_split, nccl_send, nccl_recv
     from cupy.cuda.nccl import NcclCommunicator
 else:
     NcclCommunicator = Any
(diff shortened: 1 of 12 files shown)


tharittk:
I see, for the third point about passing `base_comm_nccl=x.base_comm_nccl`, currently I test on calling `Op @ x ` which calls the concrete instance of `MPIVStack` and thus `nccl_comm` is passed (even when I did not change the `MPILinearOperator` interface). So when I check which  collective calls y operates, it says NCCL.  

If somehow the `Op @ x` is called and `Op is an instance of `MPILinearOperator`, `nccl_comm` is lost because `y` now takes the default value of `base_comm_nccl=None`.That is something I did not catch.

tharittk:
The most recent commits reflect some changes I want to point out
 - Remove of `base_comm_nccl` argument to the concrete `MPILinearOperator` class. The argument seems unnecessary and such information can be taken from operand `x` whenever math `_matvec` or `_rmatvec` is called.
 - adding of `nccl_send` and `nccl_recv` - to test this implementation, the test case of `MPIVStack` is not enough. it requires to have a test that `cell_front` is not zero i.e., not sending empty buffer. So I add the `test_blockdiag_nccl.py`. In this file, the `StackedBDiag` has the`FirstDerivative` called. This first derivative triggners the `nccl_send` and `nccl_recv` with meaning cells.

pylops_mpi/DistributedArray.py
```diff
@@ -503,6 +503,35 @@ def _allgather(self, send_buf, recv_buf=None):
             self.base_comm.Allgather(send_buf, recv_buf)
             return recv_buf
 
+    def _send(self, send_buf, dest, count=None, tag=None):
+        """ Send operation
+        """
+        if deps.nccl_enabled and getattr(self, "base_comm_nccl"):
```
mrava87, line 509: Isn't enough to use `self.base_comm_nccl`

pylops_mpi/DistributedArray.py
```diff
@@ -503,6 +503,35 @@ def _allgather(self, send_buf, recv_buf=None):
...
+                count = send_buf.size
+            nccl_send(self.base_comm_nccl, send_buf, dest, count)
+        else:
+            self.base_comm.Send(send_buf, dest, tag)
+
+    def _recv(self, recv_buf=None, source=0, count=None, tag=None):
+        """ Receive operation
+        """
+        # NCCL must be called with recv_buf. Size cannot be inferred from
+        # other arguments and thus cannot dynamically allocated
```
mrava87, line 521: cannot  -> cannot be

pylops_mpi/DistributedArray.py
```diff
@@ -503,6 +503,35 @@ def _allgather(self, send_buf, recv_buf=None):
...
+        if deps.nccl_enabled and getattr(self, "base_comm_nccl") and recv_buf is not None:
```
mrava87, line 522: mmh what if `deps.nccl_enabled and getattr(self, "base_comm_nccl")` is True but one is by mistake not passing `recv_buf`? So far I think it will go to the MPI part.... instead I think we should raise an error

pylops_mpi/DistributedArray.py
```diff
@@ -747,38 +776,55 @@ def add_ghost_cells(self, cells_front: Optional[int] = None,
 
         """
         ghosted_array = self.local_array.copy()
+        ncp = get_module(getattr(self, "engine"))
```
mrava87, line 779: why not just `self.engine`?

pylops_mpi/DistributedArray.py
```diff
@@ -747,38 +776,55 @@ def add_ghost_cells(self, cells_front: Optional[int] = None,
...
             if self.rank != 0:
-                ghosted_array = np.concatenate([self.base_comm.recv(source=self.rank - 1, tag=1), ghosted_array],
-                                               axis=self.axis)
-            if self.rank != self.size - 1:
+                # from receiver's perspective (rank), the recv buffer have the same shape as the sender's array (rank-1)
+                # in every dimension except the shape at axis=self.axis
+                recv_shape = list(recv_shapes[self.rank - 1])
+                recv_shape[self.axis] = total_cells_front[self.rank]
+                recv_buf = ncp.zeros(recv_shape, dtype=ghosted_array.dtype)
+                # Some communication can skip if len(recv_buf) = 0
```
mrava87, line 793: Some communication can skip -> Transfer of ghost cells can be skipped

pylops_mpi/utils/_nccl.py
```diff
@@ -286,3 +288,57 @@ def nccl_asarray(nccl_comm, local_array, local_shapes, axis) -> cp.ndarray:
...
+    nccl_comm : :obj:`cupy.cuda.nccl.NcclCommunicator`
+        The NCCL communicator used for point-to-point communication.
+    send_buf : :obj:`cupy.ndarray`
+        The array containing data to send.
+    dest: :obj:`int`
+        The rank of the destination GPU device.
+    count : :obj:`int`
+        Number of elements to send from `send_buf`.
+
+    Returns
```
mrava87, line 308: If nothing is returned, just not add the `Returns` section

pylops_mpi/utils/_nccl.py
```diff
@@ -4,7 +4,9 @@
     "nccl_allgather",
```
mrava87, line 4: I think it may be good to add all `nccl_*` methods to the Utils section of https://github.com/PyLops/pylops-mpi/blob/main/docs/source/api/index.rst

mrava87 (commented):
@tharittk very good! 

Everything makes sense to me and I agree with the code changes. I just left some minor suggestions. After you have taken those into account, I think we can merge this PR 😄 

One more minor thing, as you progress don't forget to keep the table in the `gpu.rst` up to date.

pylops_mpi/utils/_nccl.py
tharittk: Alright, I can do that. Maybe in the other PR ?

pylops_mpi/DistributedArray.py
tharittk: Actually, I think you mean retaining `deps.nccl_enabled` so it is
`deps.nccl_enabled and self.base_comm_nccl` right ? In this way, user can run the same code with NCCL turn off if they want to use pure MPI.

mrava87: Oh ok, makes sense 😄

mrava87 (approved):
@tharittk I have reviewed the new additions and they look great to me!

There are still a few conversations to resolve (for one of them we can hopefully get @rohanbabbar04's opinion), and I would like to hear from @rohanbabbar04 if he has any general comment, after that I will merge.

tharittk:
Thanks for the review @rohanbabbar04  @mrava87  ! 
I have pushed latest commit to reflect the suggested changes.

rohanbabbar04 (approved):

PR #139, opened 2025-06-16
feat: added masked option to asarray

mrava87:
This PR adds a new optional input to the `DistributedArray.asarray` method called `masked`, which can be used with distributed arrays that have a `mask` to create a global view of the distributed array that contains only elements from one color of the mask (so that repeated portions of the array are not returned).

### DIFF
--- a/examples/plot_distributed_array.py
+++ b/examples/plot_distributed_array.py
@@ -140,6 +140,11 @@ x = pylops_mpi.DistributedArray(global_shape=global_shape,
                                 mask=mask)
 x[:] = (MPI.COMM_WORLD.Get_rank() % subsize + 1.) * np.ones(local_shape)
 xloc = x.asarray()
+xloc1 = x.asarray(masked=True)
+
+if rank == 0:
+    print('xloc (with repeated portions)', xloc)
+    print('xloc (only effective portions):', xloc1)
 
 # Dot product
 dot = x.dot(x)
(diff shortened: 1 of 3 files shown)


rohanbabbar04 (approved):
Looks good to me!!

tharittk:
I have tested it by copy-paste your update. This uncovers the bug in `nccl_asarray()` function when working with subcomm. I am not sure exactly where yet but it looks like current `asarray()` will call `nccl_asarray()` with `self.local_shapes` (which call all GPUs in the world communicator) and may mess with the subcomm. 

The MPI does not have this problem because it sends the object and no buffer / array unrolling needed.

mrava87:
> I have tested it by copy-paste your update. This uncovers the bug in `nccl_asarray()` function when working with subcomm. I am not sure exactly where yet but it looks like current `asarray()` will call `nccl_asarray()` with `self.local_shapes` (which call all GPUs in the world communicator) and may mess with the subcomm.
> 
> The MPI does not have this problem because it sends the object and no buffer / array unrolling needed.

@tharittk thanks for checking! I am not sure I understand what the problem is with NCCL but I agree it may be due to `self.local_shapes` that is not in the subcomm but in the total comm... I suggest that I merge this as is and make a Github Issue pointing to the problem with NCCL so you can handle it in a separate PR adding also tests and @astroC86 can use these additions in his PR (which targets only MPI so far)

Issue #140, opened 2025-06-17
asarray(masked=True) does not work with NCCL


mrava87:
PR https://github.com/PyLops/pylops-mpi/pull/139 introduces the `masked` parameter to `asarray` to be able to collect a masked array from all ranks where the mask is the one given when initializing the `DistributedArray`. 

@tharittk discovered an issue with the NCCL backend likely due to `self.local_shapes` (which call all GPUs in the world communicator). MPI does not have this problem because it sends the object and no buffer / array unrolling needed.

PR #142, opened 2025-06-18
fix nccl subcommunicator bug with asarray()

tharittk:
In the issue #140, the new test uncovers a bug from NCCL when used with parititioned communicators (sub-communicator).

The issue arises because the NCCL collective calls need the recv buffer and subsequent unrolling after `allgather` (while MPI can send arbitary Python object and pack them in a list)

Fix:
- The recv buffer allocation was previously implemented as having size `MPI.COMM_WORLD.Get_size() * send_buf.size` in `_nccl.py` This will break in case of sub-communicators. 

- Manual unrolling of `local_shapes` in case of sub-communicators after `_allgather_subcomm()`

### DIFF
--- a/pylops_mpi/DistributedArray.py
+++ b/pylops_mpi/DistributedArray.py
@@ -340,7 +340,7 @@ class DistributedArray:
         local_shapes : :obj:`list`
         """
         if deps.nccl_enabled and getattr(self, "base_comm_nccl"):
-            return self._nccl_local_shapes(self.base_comm_nccl)
+            return self._nccl_local_shapes(False)
         else:
             return self._allgather(self.local_shape)
 
@@ -375,9 +375,8 @@ class DistributedArray:
             return self.local_array
 
         if deps.nccl_enabled and getattr(self, "base_comm_nccl"):
-            local_shapes = self._nccl_local_shapes(self.sub_comm if masked else self.base_comm_nccl)
             return nccl_asarray(self.sub_comm if masked else self.base_comm_nccl,
-                                self.local_array, local_shapes, self.axis)
+                                self.local_array, self._nccl_local_shapes(masked), self.axis)
         else:
             # Gather all the local arrays and apply concatenation.
             if masked:
@@ -550,14 +549,13 @@ class DistributedArray:
             self.base_comm.Recv(buf=recv_buf, source=source, tag=tag)
             return recv_buf
 
-    def _nccl_local_shapes(self, nccl_comm: NcclCommunicatorType):
+    def _nccl_local_shapes(self, masked: bool):
         """Get the the list of shapes of every GPU in the communicator
         """
         # gather tuple of shapes from every rank within thee communicator and copy from GPU to CPU
-        if nccl_comm == self.sub_comm:
+        if masked:
             all_tuples = self._allgather_subcomm(self.local_shape).get()
         else:
-            assert (nccl_comm == self.base_comm_nccl)
             all_tuples = self._allgather(self.local_shape).get()
         # NCCL returns the flat array that packs every tuple as 1-dimensional array
         # unpack each tuple from each rank


pylops_mpi/DistributedArray.py
```diff
@@ -380,8 +380,14 @@ def asarray(self, masked: bool = False):
             return self.local_array
 
         if deps.nccl_enabled and getattr(self, "base_comm_nccl"):
-            return nccl_asarray(self.sub_comm if masked else self.base_comm,
-                                self.local_array, self.local_shapes, self.axis)
+            if masked:
+                all_tuples = self._allgather_subcomm(self.local_shape).get()
```
mrava87, line 384: So I assume this is done via NCCL for how `self._allgather_subcomm` is implemented? And it looks very similar this this https://github.com/tharittk/pylops-mpi/blob/45bee369b2e4dd92eb93571ce7aa67a81bc8f5ef/pylops_mpi/DistributedArray.py#L344?

Maybe worth making a `_nncl_local_shapes` private method that is given a comm and does the packing of all the local shapes in the comm?

pylops_mpi/utils/_nccl.py
```diff
@@ -149,7 +149,7 @@ def nccl_allgather(nccl_comm, send_buf, recv_buf=None) -> cp.ndarray:
     )
     if recv_buf is None:
         recv_buf = cp.zeros(
-            MPI.COMM_WORLD.Get_size() * send_buf.size,
+            nccl_comm.size() * send_buf.size,
```
mrava87, line 152: I guess this was a bug?

mrava87 (approved):
Very good! I just left a comment for a bit of code that seems to be now repeated in two places, may be good to wrap it into a function... I leave it to you to choose if you'd like to do it here or later on. Just let me know if you want me to merge this PR 😄

pylops_mpi/utils/_nccl.py
tharittk: Yes, this is the one.

pylops_mpi/DistributedArray.py
```diff
@@ -560,6 +550,21 @@ def _recv(self, recv_buf=None, source=0, count=None, tag=None):
...
             return recv_buf
 
+    def _nccl_local_shapes(self, nccl_comm: NcclCommunicatorType):
+        """Get the the list of shapes of every GPU in the communicator
+        """
+        # gather tuple of shapes from every rank within thee communicator and copy from GPU to CPU
+        if nccl_comm == self.sub_comm:
+            all_tuples = self._allgather_subcomm(self.local_shape).get()
+        else:
+            assert (nccl_comm == self.base_comm_nccl)
```
mrava87, line 560: I do not like asserts in code, they should only be used in tests... why not passing comm and masked and then have the `if masked...` as in the code you deleted, so we don't do twice the same checks? 

I am also not so sure why before we had
```
else:
    local_shapes = self.local_shapes
```
but now also for the case without subcomm we repeat the creation of `local_shapes`

mrava87 (changes requested):
@tharittk I left some comments, I am not so convinced about the code... feels like you are repeating some operations that could be avoided (at least in the previous code you did not do them 🤔

pylops_mpi/DistributedArray.py
tharittk: Earlier, I have this
```
else:
    local_shapes = self.local_shapes
```
so that I can have single return statement `return local_shapes` , but that may look unnecessary.
Now I change the code to look like this:
```
    def _nccl_local_shapes(self, masked: bool):
        if masked:
            all_tuples = self._allgather_subcomm(self.local_shape).get()
        else:
            all_tuples = self._allgather(self.local_shape).get()
        tuple_len = len(self.local_shape)
        local_shapes = [tuple(all_tuples[i : i + tuple_len]) for i in range(0, len(all_tuples), tuple_len)]
        return local_shapes
```
It takes only `masked: bool`. We can assume if `masked=True` then we will use subcomm and we don't have to pass it.
And then have `local_shapes()` call the `_nccl_local_shapes`
```
    @property
    def local_shapes(self):
        if deps.nccl_enabled and getattr(self, "base_comm_nccl"):
            return self._nccl_local_shapes(False)
        else:
            return self._allgather(self.local_shape)
```
It is because I want to have this unpacking appear only in one place.
```
        tuple_len = len(self.local_shape)
        local_shapes = [tuple(all_tuples[i : i + tuple_len]) for i in range(0, len(all_tuples), tuple_len)]
```

mrava87 (approved):
@tharittk perfect, this version looks great 😄

PR #143, opened 2025-06-21
Fix sphinx warnings

rohanbabbar04:
- Fixing some Sphinx warnings which can be avoided.


mrava87 (approved):
Good stuff, thanks for fixing the warnings 😀

PR #146, opened 2025-06-24
doc: added tutorials with cupy+mpi

mrava87:
This PR adds a new directory called `tutorials_cupy` which includes tutorials running with CuPy arrays and MPI communicators. 

Following what previously done for `tutorials_nccl`, a couple of new targets are also added to the `Makefile` to allow building the documentation with MPI or MPI/Cupy+MPI or MPI/CuPy+MPI/CuPy+NCCL tutorials.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,10 @@ PIP := $(shell command -v pip3 2> /dev/null || command which pip 2> /dev/null)
 PYTHON := $(shell command -v python3 2> /dev/null || command which python 2> /dev/null)
 NUM_PROCESSES = 3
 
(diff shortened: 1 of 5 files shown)


rohanbabbar04 (commented):
Using `rank % device_count` seems appropriate, as it allows flexibility in the number of ranks without requiring an equal number of GPUs.

mrava87:
> Using `rank % device_count` seems appropriate, as it allows flexibility in the number of ranks without requiring an equal number of GPUs.
> 
> Otherwise, it looks great—let me know if you agree with me 🙂

Thank @rohanbabbar04! Yes, I agree, that is a good point 😄 I am going to do give quick look at all the tutorials and then merge

rohanbabbar04 (approved):

PR #150, opened 2025-07-01
test: fix tests not working with high nranks and change GA to test that

mrava87:
This PR fixes a number issues that arose when trying to increase the number of ranks in the test suite to 9 or 16. 

As a consquence, I have also decided to modify the GA `build` to run tests with 2,4,9 ranks instead of 2,3,4

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -20,6 +20,8 @@ jobs:
         exclude:
           - os: macos-latest
             mpi: 'intelmpi'
+          - mpi: 'openmpi'
+            rank: '9' # works locally, not in CI
     runs-on: ${{ matrix.os }}
     steps:
       - name: Checkout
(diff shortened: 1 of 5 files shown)


mrava87:
@rohanbabbar04 no idea why the `openmpi` runs with N=9 stall, but those with `intelmpi` and `mpich` don't... note that without the proposed changes some of the tests failed for N=9 for me locally... most are tests I added with the `mask` and add some clear if/else statement on the number of ranks and did not stop if a different one was passed, and another one was for the solvers due to some stochasticity in the random (if I run it independently, instead as part of the entire test suite, they were passing for N=9)

rohanbabbar04:
Hmm, let me give it a look

rohanbabbar04 (approved):
Just added a minor change to the OpenMPI execution by using `^openib` to exclude InfiniBand, since GitHub Actions runners may stall/show error messages when running with higher ranks.

I will now merge this :slightly_smiling_face:

PR #151, opened 2025-07-01
test: started adding tests for nccl utility routines

mrava87:
This is a WIP PR to add tests for various `nccl` utility routines in `pylops_mpi.utils._nccl` for primitive communication calls.

The need for this tests stems from the realization that if we pass `send_buf` of different size to `nccl_allgather` we get a deadlock... so we should make sure to check and raise an error before this happens.

Note that the last test is currently commented out because mpiexec will hang if running this test - I believe this is due to the fact that `nccl_comm.allGather` does not cope with `send_buf` of different size between ranks


mrava87:
Superseded by https://github.com/PyLops/pylops-mpi/pull/148/files#diff-6a32c7fc495c6e1e40bfcc419fd495e484556369befae6a493a25b078e809736

PR #155, opened 2025-07-03
test/doc: added tests and example of chaining MPIMatrixMult

mrava87:
This PR introduces an example and some tests of chaining the new `MPIMatrixMult` operator with a block diagonal operator that applies a first derivative to the different column blocks of the output of the distributed matrix multiplication 

@astroC86, let me know what you think 😄


astroC86:
looks fine, I now also see what you meant in the previous PR.

PR #157, opened 2025-07-11
Adding Benchmark instrument

tharittk:
This PR introduces a `benchmark` decorator that enables users to place `mark(label)` calls anywhere inside a function to record execution time at specific code points. The timing marks are collected and (as of now) printed in a human-readable format after the function completes.

The design is inspired by C++-style instrumentation, where developers can inject performance markers directly into the code.

Changes made: 
- `benchmark.py` A benchmark decorator for lightweight instrumentation

- `kirchhoff_bench.py` A sample script to demonstrate how to use the decorator

Ongoing work:
- [ ] Develop a more thorough benchmarking script with varying:
Model/data sizes (e.g., 1 KB → 8 MB).
Number of GPU nodes

- [ ] Extend benchmarking to cover:
MDD (Multi-Dimensional Deconvolution).
Communication primitives like all_gather, all_reduce, etc.

### DIFF
--- a/pylops_mpi/utils/benchmark.py
+++ b/pylops_mpi/utils/benchmark.py
@@ -1,0 +1,62 @@
+import functools
+import time
+
+# TODO (tharitt): later move to env file or something
+ENABLE_BENCHMARK = True
+
+# This function is to be instrumented throughout the targeted function
+def mark(label):
+    if _current_mark_func is not None:
+        _current_mark_func(label)
+
+# Global hook - this will be re-assigned (points to)
+# the function defined in benchmark wrapper
+_current_mark_func = None
+
+def benchmark(func):
+    """A wrapper for code injection for time measurement.
+
+    This wrapper allows users to put a call to mark()
+    anywhere inside the wrapped function. The function mark()
+    is defined in the global scope to be a placeholder for the targeted
+    function to import. This wrapper will make it points to local_mark() defined
+    in this function. Therefore, the wrapped function will be able call 
+    local_mark(). All the context for local_mark() like mark list can be
+    hidden from users and thus provide clean interface.
+
+    Parameters
+    ----------
+    func : :obj:`callable`, optional
+        Function to be decorated.
+    """
+
+    # Zero-overhead
+    if not ENABLE_BENCHMARK:
+        return func
+
+    @functools.wraps(func)
+    def wrapper(*args, **kwargs):
+        marks = []
+
+        # currently this simply record the user-define label and record time
+        def local_mark(label):
+            marks.append((label, time.perf_counter()))
+
+        global _current_mark_func
+        _current_mark_func = local_mark
+
+        # the mark() called in wrapped function will now call local_mark
+        result = func(*args, **kwargs)
+        # clean up to original state
+        _current_mark_func = None
+
+        # TODO (tharitt): maybe changing to saving results to file instead 
+        if marks:
+            prev_label, prev_t = marks[0]
+            print(f"[BENCH] {prev_label}: 0.000000s")
+            for label, t in marks[1:]:
+                print(f"[BENCH] {label}: {t - prev_t:.6f}s since '{prev_label}'")
+                prev_label, prev_t = label, t
+        return result
+
+    return wrapper
--- a/pylops_mpi/utils/benchmark.py
+++ b/pylops_mpi/utils/benchmark.py
@@ -4,14 +4,17 @@ import time
 # TODO (tharitt): later move to env file or something
 ENABLE_BENCHMARK = True
 
+
 # This function is to be instrumented throughout the targeted function
 def mark(label):
     if _current_mark_func is not None:
         _current_mark_func(label)
 
+
 # Global hook - this will be re-assigned (points to)
 # the function defined in benchmark wrapper
 _current_mark_func = None
+
 
 def benchmark(func):
     """A wrapper for code injection for time measurement.
@@ -20,7 +23,7 @@ def benchmark(func):
     anywhere inside the wrapped function. The function mark()
     is defined in the global scope to be a placeholder for the targeted
     function to import. This wrapper will make it points to local_mark() defined
-    in this function. Therefore, the wrapped function will be able call 
+    in this function. Therefore, the wrapped function will be able call
     local_mark(). All the context for local_mark() like mark list can be
     hidden from users and thus provide clean interface.
 
@@ -50,7 +53,7 @@ def benchmark(func):
         # clean up to original state
         _current_mark_func = None
 
-        # TODO (tharitt): maybe changing to saving results to file instead 
+        # TODO (tharitt): maybe changing to saving results to file instead
         if marks:
             prev_label, prev_t = marks[0]
             print(f"[BENCH] {prev_label}: 0.000000s")
--- a/pylops_mpi/benchmark/kirchhoff_bench.py
+++ b/pylops_mpi/benchmark/kirchhoff_bench.py
@@ -1,0 +1,107 @@
+import warnings
+warnings.filterwarnings('ignore')
+
+import numpy as np
+import cupy as cp
+from pylops_mpi.utils.benchmark import benchmark, mark
+from mpi4py import MPI
+
+from pylops.utils.wavelets import ricker
+from pylops.waveeqprocessing.lsm import LSM
+
+import pylops_mpi
+np.random.seed(42)
+rank = MPI.COMM_WORLD.Get_rank()
+
+# Trade-off: model size is controlled by (nx, nz) which affects data transfer and
+# (nr, ns) controls data size which affects the compute
+par1 = {"nx": 30, "nz": 50, "ns": 9, "nr": 18, "use_cupy": False, "use_nccl": False}
+par2 = {"nx": 30, "nz": 50, "ns": 9, "nr": 18, "use_cupy": True, "use_nccl": False}
+par3 = {"nx": 30, "nz": 50, "ns": 9, "nr": 18, "use_cupy": True, "use_nccl": True}
+
+
+def prepare_kirchhoff_op(par):
+    v0 = 1500
+    dx = 12.5
+    dz = 4
+    x, z = np.arange(par["nx"]) * dx, np.arange(par["nz"]) * dz
+
+    # recv and source config
+    rx = np.linspace(10 * dx, (par["nx"] - 10) * dx, par["nr"])
+    rz = 20 * np.ones(par["nr"])
+    recs = np.vstack((rx, rz))
+
+    nstot = MPI.COMM_WORLD.allreduce(par["ns"], op=MPI.SUM)
+    sxtot = np.linspace(dx * 10, (par["nx"] - 10) * dx, nstot)
+    sx = sxtot[rank * par["ns"]: (rank + 1) * par["ns"]]
+    sz = 10 * np.ones(par["ns"])
+    sources = np.vstack((sx, sz))
+
+    # Wavelet
+    nt = 651
+    dt = 0.004
+    t = np.arange(nt) * dt
+    wav, wavt, wavc = ricker(t[:41], f0=20)
+
+    lsm_op = LSM(
+        z,
+        x,
+        t,
+        sources,
+        recs,
+        v0,
+        cp.asarray(wav.astype(np.float32)) if par["use_cupy"] else wav,
+        wavc,
+        mode="analytic",
+        engine="cuda" if par["use_cupy"] else "numba",
+        dtype=np.float32
+    )
+    if par["use_cupy"]:
+        lsm_op.Demop.trav_srcs = cp.asarray(lsm_op.Demop.trav_srcs.astype(np.float32))
+        lsm_op.Demop.trav_recs = cp.asarray(lsm_op.Demop.trav_recs.astype(np.float32))
+
+    return lsm_op
+
+
+def prepare_distributed_data(par, lsm_op, nccl_comm):
+    # Reflectivity Model
+    refl = np.zeros((par["nx"], par["nz"]))
+    refl[:, par["nz"] // 4] = -1
+    refl[:, par["nz"] // 2] = 0.5
+    refl_dist = pylops_mpi.DistributedArray(global_shape=par["nx"] * par["nz"],
+                                            partition=pylops_mpi.Partition.BROADCAST,
+                                            base_comm_nccl=nccl_comm,
+                                            engine="cupy" if par["use_cupy"] else "numpy")
+    refl_dist[:] = cp.asarray(refl.flatten()) if par["use_cupy"] else refl.flatten()
+
+    VStack = pylops_mpi.MPIVStack(ops=[lsm_op.Demop, ])
+    d_dist = VStack @ refl_dist
+    return d_dist
+
+
+@benchmark
+def run_bench(par):
+    # if run with MPI, NCCL should not be initialized at all to avoid hang
+    if par["use_nccl"]:
+        nccl_comm = pylops_mpi.utils._nccl.initialize_nccl_comm()
+    else:
+        nccl_comm = None
+
+    mark(f"begin {par["use_cupy"]=}, {par["use_nccl"]=}")
+    lsm_op = prepare_kirchhoff_op(par)
+    d_dist = prepare_distributed_data(par, lsm_op, nccl_comm)
+    VStack = pylops_mpi.MPIVStack(ops=[lsm_op.Demop, ])
+    mark("after prepare")
+    # TODO (tharitt): In the actual benchmark, we probably have to decorate
+    # the matvec() and rmatvec() to separate computation from communication time
+    madj_dist = VStack.H @ d_dist
+    mark("after adjoint")
+    _ = madj_dist.asarray().reshape((par["nx"], par["nz"]))
+
+
+if __name__ == "__main__":
+    run_bench(par1)
+    print("========")
+    run_bench(par2)
+    print("========")
+    run_bench(par3)


mrava87:
@tharittk I’ll take a look at this soon. Regarding the 2 points of ongoing work, I think these kind of extensive benchmark tests do not belong to the main repo but they are better suited for the one we created to put test codes… here we should add some minimal tests to verify the correctness of the benchmark decorator and mark

mrava87 (commented):
@tharittk good start! I think we will probably need a few iterations here to get something that is functional and looks nice, and I am playing myself with it to get a better feeling 😄 

Here are some initial comments as they come:
- definitely we should not have benchmark scripts in the main repo, so `Kirchhoff_bench` has to be ultimately removed from here and put in separate repo
- we should try to add a new page to the documentation called `Benchmarking` where we can provide an explanation of this decorator, how to use it, and show a small example (a bit like we have in the `GPU Support` and `Adding New Operators` pages)
- I tried to add a benchmark decorator to `nccl_allreduce` but what I see is that the messages of benchmark come all before those of mark.. is this expected/intended?
- In general I am never in favour of re-inventing the wheel.. so i have been wondering if we really need to write code to print/write on file and after a few tests I think going for the route of using `logging` is probably better for the long run (we can get lots of things for free including writing to file. A little example:

In `kirchhoff_bench`:

```
import logging
logging.basicConfig(level=logging.INFO, force=True) or logging.basicConfig(filename='kirch_log.log', level=logging.INFO, force=True)
numba_logger = logging.getLogger('numba') 
numba_logger.setLevel(logging.WARNING) # silence numba (can do any other library we get logs one doesn't want to see)

```

and in `benchmark`

```
import logging
logger = logging.getLogger(__name__)

... 
print("".join(output)) -> logging.info("".join(output))
```

pylops_mpi/utils/benchmark.py
```diff
@@ -0,0 +1,99 @@
...
+import time
+from typing import Callable, Optional
+from mpi4py import MPI
+
+
+# TODO (tharitt): later move to env file or something
+ENABLE_BENCHMARK = True
+
+
+# This function allows users to measure time arbitary lines of the function
```
mrava87, line 11: Move this to a proper docstring

pylops_mpi/utils/benchmark.py
```diff
@@ -0,0 +1,99 @@
...
+def mark(label):
+    if not _mark_func_stack:
+        raise RuntimeError("mark() called outside of a benchmarked region")
+    _mark_func_stack[-1](label)
+
+
+# Stack of active mark functions for nested support
```
mrava87, line 18: I would move the global variables above function definitons

pylops_mpi/utils/benchmark.py
```diff
@@ -0,0 +1,99 @@
...
+            _mark_func_stack.append(local_mark)
+
+            start_time = time.perf_counter()
+            # the mark() called in wrapped function will now call local_mark
+            result = func(*args, **kwargs)
+            end_time = time.perf_counter()
+
+            _mark_func_stack.pop()
+
+            output = []
```
mrava87, line 74: I'd move output just before the `if rank` as that should not be time (though it won't affect timing its cleaner to have it after..)

pylops_mpi/utils/benchmark.py
```diff
@@ -0,0 +1,99 @@
...
+            def local_mark(label):
+                marks.append((label, time.perf_counter()))
+            _mark_func_stack.append(local_mark)
+
+            start_time = time.perf_counter()
+            # the mark() called in wrapped function will now call local_mark
+            result = func(*args, **kwargs)
+            end_time = time.perf_counter()
+
+            _mark_func_stack.pop()
```
mrava87, line 72: add a short comment why this is done

pylops_mpi/utils/benchmark.py
```diff
@@ -0,0 +1,99 @@
...
+
+            output = []
+            # start-to-end time
+            elapsed = end_time - start_time
+
+            # TODO (tharitt): Both MPI + NCCL collective calls have implicit synchronization inside the stream
+            # So, output only from rank=0 should suffice. We can add per-rank output later on if makes sense.
+            if rank == 0:
```
mrava87, line 80: I see the point about rank0, as a start i think it is fine (especially once we add the syncronization timings would not be different in other ranks).. but then its no point to have ` (rank = {rank})\n")` in the text to print

pylops_mpi/utils/benchmark.py
```diff
@@ -0,0 +1,99 @@
...
+                level = len(_mark_func_stack)
+                output.append(f"{'---' * level}{description or func.__name__} {elapsed:6f} seconds (rank = {rank})\n")
```
mrava87, line 82: Not sure i get this `"{'---' * level}`, is it for recognising nesting?

mrava87: Also add `:` before elapsed to be consistent with what is done in the mark output strings

mrava87: and either s here or seconds after

pylops_mpi/utils/benchmark.py
```diff
@@ -0,0 +1,99 @@
+import functools
+import time
+from typing import Callable, Optional
+from mpi4py import MPI
+
+
+# TODO (tharitt): later move to env file or something
```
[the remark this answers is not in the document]
tharittk: Got it.

pylops_mpi/utils/benchmark.py
tharittk: agree with that

pylops_mpi/utils/benchmark.py
tharittk: Yeah the `"{'---' * level}` is to signify the nesting. The depth of the nesting equals the length of the `_mark_fun_stack`. 

This is still not quite clean I think. As in this example (after fixing those allreduce calls from `_check_local_shapes` to use MPI instead of NCCL)
```
$ srun -n 3 python kirchhoff_bench.py 
---nccl_allreduce 0.000104 seconds 

run_bench 2.208081 seconds 
[begin ---> done initialize]: 1.199786s 
[done initilize ---> done prepare]: 0.415888s 
[done prepare ---> done adjoint]: 0.093718s 
```

Ideally, the ---nccl_allreduce 0.000104 seconds shoud go directly below the [done prepare --> done adjoin] like this 
```
[done prepare ---> done adjoint]: 0.093718s
---nccl_allreduce 0.000104 seconds 
```

tharittk: Ok think I got the better implementation now. The output looks like this. I recognize that the nested calls are the pre-order traversal of the function calls. So you collect all these with info of level. Then parse these output later at the end. I will push the commit later today.

```
[header]run_bench: total runtime: 1.721294
        begin (NCCL init)-->Start prepare Kirchh:  1.210365
        Start prepare Kirchh-->Start Adjoint Op:  0.385340
        [header]VStack rmatvec: total runtime: 0.095414
                [header]nccl_allreduce: total runtime: 0.000125
                        nccl_prepare-->nccl done prepare:  0.000040
                        nccl done prepare-->nccl done allreduce:  0.000080
        Start Adjoint Op-->Done Adjoint Op:  0.095447
```

pylops_mpi/utils/benchmark.py
```diff
@@ -0,0 +1,99 @@
...
+
+            # TODO (tharitt): Both MPI + NCCL collective calls have implicit synchronization inside the stream
+            # So, output only from rank=0 should suffice. We can add per-rank output later on if makes sense.
+            if rank == 0:
+                level = len(_mark_func_stack)
+                output.append(f"{'---' * level}{description or func.__name__} {elapsed:6f} seconds (rank = {rank})\n")
+                if marks:
+                    prev_label, prev_t = marks[0]
+                    for label, t in marks[1:]:
+                        output.append(f"{'---' * level}[{prev_label} ---> {label}]: {t - prev_t:.6f}s \n")
```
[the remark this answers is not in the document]
tharittk: Yes, I'll handle this as mentioned in the comment above. I will look something like this
```
[header]run_bench: total runtime: 1.721294
        begin (NCCL init)-->Start prepare Kirchh:  1.210365
        Start prepare Kirchh-->Start Adjoint Op:  0.385340
        [header]VStack rmatvec: total runtime: 0.095414
                [header]nccl_allreduce: total runtime: 0.000125
                        nccl_prepare-->nccl done prepare:  0.000040
                        nccl done prepare-->nccl done allreduce:  0.000080
        Start Adjoint Op-->Done Adjoint Op:  0.095447
```

pylops_mpi/utils/benchmark.py
```diff
@@ -0,0 +1,145 @@
...
+              save_file=False,
+              file_path='benchmark.log'
+              ):
+    """A wrapper for code injection for time measurement.
+
+    This wrapper measure the start-to-end time of the wrapped function when
+    decorated without any argument.
+    It also allows users to put a call to mark() anywhere inside the wrapped function
+    for fine-grain time benchmark. This wrapper defines the local_mark() and push it
+    the the _mark_func_stack for isolation in case of nested call.
```
mrava87, line 70: the the  ->   to the

pylops_mpi/utils/benchmark.py
```diff
@@ -0,0 +1,145 @@
...
+
+            _mark_func_stack.append(local_mark)
+
+            start_time = time.perf_counter()
+            # the mark() called in wrapped function will now call local_mark
+            result = func(*args, **kwargs)
+            end_time = time.perf_counter()
+
+            elapsed = end_time - start_time
+            _markers[header_index] = (f"[header]{description or func.__name__}", elapsed, level)
```
mrava87, line 112: What does `[header]` mean? Maybe it would make more sense to put something like `[decorator]`?

mrava87 (changes requested):
@tharittk well done, I like the new formatting with messages from nested benchmark appearing in the right order 😄 

I left a few additional comments/thoughts for you... once you have addressed the remaining points that you have in the PR main comment, I'll do a final review

pylops_mpi/utils/benchmark.py
tharittk: For lack of a better word, I use the "header" to signify that this is the function decorated by @benchmark. Other lines that measure time come from `wrap()`.

I agree we can change it to [decorator] or [decorated]. It is more precise.

pylops_mpi/utils/benchmark.py
```diff
@@ -0,0 +1,145 @@
...
+        A label of the mark. This signifies both 1) the end of the
+        previous mark 2) the beginning of the new mark
+    """
+    if not _mark_func_stack:
+        raise RuntimeError("mark() called outside of a benchmarked region")
+    _mark_func_stack[-1](label)
+
+
+def benchmark(func: Optional[Callable] = None,
+              description="",
```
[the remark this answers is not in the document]
tharittk: Oh thanks for catching that

pylops_mpi/utils/benchmark.py
```diff
@@ -0,0 +1,145 @@
...
+import time
+from typing import Callable, Optional
+from mpi4py import MPI
+
+
+# TODO (tharitt): later move to env file or something
+ENABLE_BENCHMARK = True
+
+logger = logging.getLogger(__name__)
+logging.basicConfig(level=logging.INFO, force=True)
```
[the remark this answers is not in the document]
tharittk: Yes, this is to ensure that any previous handlers are closed. However, I am not sure this is the proper way. The reason I have `force=True` is simply to be sure that I isolate this benchmark logging and get a clean start.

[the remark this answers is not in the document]
tharittk: Actually, after making the change that you suggested below to remove the logger configuration,  these parts can be removed entirely.

pylops_mpi/utils/benchmark.py
```diff
@@ -1,24 +1,28 @@
...
-
 # Stack of active mark functions for nested support
 _mark_func_stack = []
 _markers = []
 
 
-def _parse_output_tree(markers: list[str]):
+def _parse_output_tree(markers: List[str]):
+    """This function parses the list of strings gathered during the benchmark call and output them
+    as one properly formatted string. The format of output string follows the hierachy of function calls
```
mrava87, line 18: hierachy -> hierarchy

mrava87 (changes requested):
Great work @tharittk ! I just left a few minor comments, but I think we have converged to a final solution 😄 

Try to handle the two remaining tasks and then we should be ready to merge this!

pylops_mpi/utils/benchmark.py
```diff
@@ -0,0 +1,99 @@
...
+            result = func(*args, **kwargs)
+            end_time = time.perf_counter()
+
+            _mark_func_stack.pop()
+
+            output = []
+            # start-to-end time
+            elapsed = end_time - start_time
+
+            # TODO (tharitt): Both MPI + NCCL collective calls have implicit synchronization inside the stream
```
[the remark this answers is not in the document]
tharittk: I was wrong on `cp.cuda.get_current_stream() synchronize ()`
If no collective call is called (like when we just do NumPy + MPI), this one will throw the error
```
  File "/u/ttangkijwanichakul/working/pylops-mpi/pylops_mpi/utils/benchmark.py", line 62, in _sync
    cp.cuda.get_current_stream().synchronize()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "cupy/cuda/stream.pyx", line 87, in cupy.cuda.stream.get_current_stream
  File "cupy/cuda/stream.pyx", line 96, in cupy.cuda.stream.get_current_stream
  File "cupy/cuda/stream.pyx", line 38, in cupy.cuda.stream._ThreadLocal.get
  File "cupy/cuda/stream.pyx", line 24, in cupy.cuda.stream._ThreadLocal.__init__
  File "cupy_backends/cuda/api/runtime.pyx", line 394, in cupy_backends.cuda.api.runtime.getDeviceCount
  File "cupy_backends/cuda/api/runtime.pyx", line 146, in cupy_backends.cuda.api.runtime.check_status
cupy_backends.cuda.api.runtime.CUDARuntimeError: cudaErrorNoDevice: no CUDA-capable device is detected
```

pylops_mpi/utils/benchmark.py
```diff
@@ -0,0 +1,145 @@
...
+                    output = _parse_output_tree(_markers)
+                    for handler in logger.handlers[:]:
+                        logger.removeHandler(handler)
+
+                    if save_file:
+                        handler = logging.FileHandler(file_path, mode='w')
+                    else:
+                        handler = logging.StreamHandler(sys.stdout)
+
+                    handler.setLevel(logging.INFO)
```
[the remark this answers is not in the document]
tharittk: Yes that's what I observed in the execution also. With the `propagate=False`, the logger only logs the message associated with this module. Previously, without the flag, I would see other message from numba-related calls.

tutorials/benchmark_example.py
```diff
@@ -0,0 +1,134 @@
+r"""
+Benchmark Utility in PyLops-MPI
+=========================
```
mrava87, line 3: same length of the title

tutorials/benchmark_example.py
```diff
@@ -0,0 +1,134 @@
...
+This tutorial demonstrates how to use the bencmark utility of PyLops-MPI. It contains various
```
mrava87, line 4: This tutorial demonstrates how to use the :py:func:`pylops_mpi.utils.benchmark` and :py:func:`pylops_mpi.utils.mark` utility methods in PyLops-MPI.

tutorials/benchmark_example.py
```diff
@@ -0,0 +1,134 @@
...
+function calling pattern that may come up during the benchmarking.
```
mrava87, line 5: during the benchmarking of a distributed code.

tutorials/benchmark_example.py
```diff
@@ -0,0 +1,134 @@
...
+                                partition=par['partition'],
+                                dtype=par['dtype'], axis=par['axis'])
+    # may perform computation here
+    dist_arr.dot(dist_arr)
+
+
+###############################################################################
+# When we call :py:func:`inner_func`, we will see the result
+# of the benchmark print to standard output. If we want to customize the
+# function name in the printout, we can pass the parameter to the :py:func:`benchmark`
```
mrava87, line 41: the parameter ``description`` to the....

tutorials/benchmark_example.py
```diff
@@ -0,0 +1,134 @@
...
+# i.e., :py:func:`@benchmark(description="printout_name")`
+
+inner_func(par)
+
+###############################################################################
+# We may want to get the finer time measurement by timing the execution time from arbitary lines
```
mrava87, line 47: the finer time measurement -> fine-grained time measurements

mrava87, line 47: from arbitary lines -> of arbitary lines

tutorials/benchmark_example.py
```diff
@@ -0,0 +1,134 @@
...
+# of code. :py:func:`pylops_mpi.utils.mark` provides such utitlity
```
mrava87, line 48: such utitlity -> such utility.

tutorials/benchmark_example.py
```diff
@@ -0,0 +1,134 @@
...
+    dist_arr = DistributedArray(global_shape=par['global_shape'],
+                                partition=par['partition'],
+                                dtype=par['dtype'], axis=par['axis'])
+    mark("Begin dot")
+    dist_arr.dot(dist_arr)
+    mark("Finish dot")
+
+
+###############################################################################
+# Now when we run, we get the detail time measurement. Noted that there is a tag
```
mrava87, line 63: detail -> detailed

mrava87: Noted that -> Note that

tutorials/benchmark_example.py
```diff
@@ -0,0 +1,134 @@
...
+# [decorator] to the function name to distinguish between the start-to-end time measuredment of
```
mrava87, line 64: to the -> next to the

mrava87: measuredment -> measurement

tutorials/benchmark_example.py
```diff
@@ -0,0 +1,134 @@
...
+# top-level function and those that comes from :py:func:`pylops_mpi.utils.mark`
```
mrava87, line 65: top -> the  top

tutorials/benchmark_example.py
```diff
@@ -0,0 +1,134 @@
...
+inner_func_with_mark(par)
+
+###############################################################################
+# This utility also supports the nested functions. Let's define the outer function
```
mrava87, line 69: This utility benchmarking routines can also be nested. Let's define an outer function that internally calls the decorated

tutorials/benchmark_example.py
```diff
@@ -0,0 +1,134 @@
...
+    dist_arr = DistributedArray(global_shape=par['global_shape'],
+                                partition=par['partition'],
+                                dtype=par['dtype'], axis=par['axis'])
+    dist_arr + dist_arr
+    mark("Outer func ends")
+
+
+###############################################################################
+# If we run :py:func:`outer_func_with_mark`, we get the time measurement nicely
+# printout with the nested indentation to specify that nested calls.
```
mrava87, line 86: printout -> printed out

tutorials/benchmark_example.py
```diff
@@ -0,0 +1,134 @@
...
+outer_func_with_mark(par)
+
+
+###############################################################################
+# In some cases, we may want to write benchmark output to a text file.
+# :py:func:`pylops_mpi.utils.benchmark` also takes the py:class:`logging.Logger`
+# in its argument. Let's first import the logging package and construct our logger
+
+import sys
```
mrava87, line 95: I would import those at the top

mrava87 (changes requested):
@tharittk very good, I like the tutorial 😄 I just left some minor comments related to typos and unclear text... 

I just noticed though that you have not yet handled this

```
# TODO (tharitt): later move to env file or something
ENABLE_BENCHMARK = True
```

which is the reason why I was asking to write a page like https://github.com/tharittk/pylops-mpi/blob/benchmark_instrument/docs/source/adding.rst more than a tutorial... note that I would maybe call it `BENCH_PYLOPS_MPI ` (always good to have the name of the library so the env variable scope is not ambiguous

I think it is fine to have the tutorial but we definitely need a page where we can replicate a bit the same info and more importantly explain how one can turn on/off benchmark; this however brings me to my last more general question: as we progressed I kind of lost a bit track of the original goal of this (lol) which is timing our communication methods in `_nccl`; at this point we have not added any decorator to them so how do you envision doing it - a developer goes it manually, decorates them when doing some benchmarking and then takes away the decorators? I guess that is a doable approach, although a bit clunky... the other way could be to decorate all of them since we have the `ENABLE_BENCHMARK /BENCH_PYLOPS_MPI` variable that is generally set to 0 (False) and that the method `benchmark` checks to decide whether to activate or not... as the way we ended up creating benchmark/mark would always log the time if we add it and print it, and that is not what we want in a normal run. 

Any thoughts?

docs/source/benchmarking.rst
mrava87, line 1: @tharittk sorry I think we did not understand each other... I did not mean to say that you need to copy-paste the tutorial, that would make no much sense.

Try to keep this page much more concise... it doesn't need to show a functioning code as we can always refer to the tutorial, e.g.

```
import sys
import logging
import numpy as np

...
```

can be all taken away... just show the key lines of code related to benchmark 😄 

Also this is the place to describe ``BENCH_PYLOPS_MPI`` which seems like is not yet being documented anywhere (and whilst do it remove 

```
TODO (tharitt): later move to env file or something
ENABLE_BENCHMARK = True
```

from the `benchmark` file.

tutorials/benchmark_example.py
mrava87, line 1: Call this just `benchmark` or `benchmarking` to be more in line with the name of the other files

mrava87 (changes requested):
@tharittk I totally agree with your judgement, let's go this way... and add the `BENCH_PYLOPS_MPI` variable that when set to 0 will skip benchmark... as you say, if unset default to 1. However, you do need to modify `benchmark.py` to handle the new env variable, which does not seem to be done yet.

But I am not super happy with the `benchmarking.rst` file yet. I left a detailed comment that hopefully can help you improve it and make it a bit more different f 😄

PR #165, opened 2025-08-03
Dual test for NumPy and CuPy in tests

tharittk:
Some bugs were found on
- Infinity norm causes segmentation fault when using with CuPy + MPI
- ? Pylops BlockDiag ** 3 seems to reverse the engine to NumPy despite being initialized as CuPy

### DIFF
--- a/docs/source/contributing.rst
+++ b/docs/source/contributing.rst
@@ -69,6 +69,18 @@ that the both old and new tests pass successfully:
 
    >> make tests
 
+If you run PyLops-MPI with GPUs you may also do:
+
+.. code-block:: bash
+
+   >> make tests_gpu
+
+Additionally, if you have a NCCL-enabled environment, you may also check:
+
+.. code-block:: bash
+
+   >> make tests_nccl
+
 4. Make sure the ``examples`` python scripts are executed using 3 processes without any errors:
 
 .. code-block:: bash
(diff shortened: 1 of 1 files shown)


mrava87:
Also I think for all tests we need to add some logic like https://github.com/PyLops/pylops-mpi/blob/b5a4c54f3270f218078d39aa384cdf128091c08a/tutorials_cupy/poststack_cupy.py#L33 to have different ranks use different GPUs otherwise they will all run on the same GPU (default=0)

tests/test_blockdiag.py
```diff
@@ -27,6 +27,14 @@
 par2j = {'ny': 301, 'nx': 101, 'dtype': np.complex128}
 
 np.random.seed(42)
+rank = MPI.COMM_WORLD.Get_rank()
```
mrava87, line 30: This looks good 😊 I’m just not sure why you need OMPI_…, is it not enough to just use rank? Or where you see that the latter would take precedence?

mrava87 (commented):
@tharittk good job!

I think this is nearly ready to go. I would maybe add some targets in the Makefile like we have in pylops https://github.com/PyLops/pylops/blob/a94ea8eae3b9c06bf39637b2e29f6a45a0e7766f/Makefile#L54 and in the contributing part of the documentation (see again what we have in pylops and maybe also add something about the NCCL tests and examples which I just realized is missing)

tests/test_blockdiag.py
tharittk: I was thinking in the case of a multi-node with multi-GPU.
If there are 2 nodes and 4 GPUs each, the rank will be 0,1,2,3, and 4,5,6,7  for node 1 and ,2 respectively.
In this case, the `device_count` should return 4 (not 8 - the size of MPI.COMM_WORLD) so everything should be OK. When I was writing code, I thought it may return 8 - and this is the reason I thought `local_rank` from MPI may be safer. I think it is possible to remove the OMPI... then.

hongyx11:
it's doable, let me give it a try, we need to use srun

PR #166, opened 2025-08-07
Implementation of distributed Halo Operator

astroC86:
This PR adds the `Halo` and `MPINonStationaryConvolve1D` operators.


pylops_mpi/basicoperators/Halo.py
```diff
@@ -0,0 +1,201 @@
...
+            raise ValueError(f"x should have partition={Partition.SCATTER} Got {x.partition} instead...")
+
+        y = DistributedArray(global_shape=self.shape[1],
+                               partition=Partition.SCATTER)
+
+        arr = x.local_array.reshape(self.local_extent)
+        core_slices = [slice(left, left + ldim) for left, ldim in zip(self.halo[::2], self.local_dims)]
+        core = arr[tuple(core_slices)]
+        y[:] = core.ravel()
+        return y
```
astroC86: add new line here

mrava87:
@astroC86 I had a first spin at the `Halo` operator. Seems very promising and I can see various things we could do with it ❤️ 

I am going to summarize my experience so far here and tomorrow I will push an updated version of `plot_halo` and a new `plot_halo1d` which uses an updated version of the actual implementation (will push it in a file called `Halo_new.py` so you can see what changes I made - so far they seem to break the 3d implementation though...):

- I tried to check if the operator passes the dottest with this very simple example, but i got `AssertionError: Dot test failed, v^H(Opu)=61.67157626475711 - u^H(Op^Hv)=65.33767302699903`... not really sure why, but it seems that the adjoint is not correct (I trust the forward as I used in various examples and it always seem to make sense)

```
import numpy as np
import pylops
from pylops_mpi.basicoperators.Halo import ScatterType, local_block_split, BoundaryType
from pylops_mpi.utils.dottest import dottest
from mpi4py import MPI

import pylops_mpi

np.random.seed(42)

comm = MPI.COMM_WORLD
rank = comm.Get_rank()
size = comm.Get_size()

# Halo operator
nlocal = 64
n = nlocal * size
proc_grid_shape = (size, )  # number of partitions over each axis

halo = 1
halo_op = pylops_mpi.basicoperators.MPIHalo(
    dims=(n, ),
    halo=halo,
    proc_grid_shape=proc_grid_shape,
    boundary_mode=BoundaryType.ZERO,
    comm=comm
)

x_dist = pylops_mpi.DistributedArray(
    global_shape=n,
    base_comm=comm,
    partition=pylops_mpi.Partition.SCATTER)
x_dist[:] = np.random.normal(0., 1., nlocal)

y_dist = pylops_mpi.DistributedArray(
    global_shape=halo_op.shape[0],
    base_comm=comm,
    partition=pylops_mpi.Partition.SCATTER)
y_dist[:] = np.random.normal(0., 1., y_dist.local_array.shape)

dottest(halo_op, x_dist, y_dist, verb=True)
```

- I used Halo to implement a distributed FirstDerivative operator using only pylops FirstDerivative and MPIBlockDiag... I think this is a nice example of how one could use Halo instead of rewrite from scratch a distributed operator and use `add_ghost_cells`.. this however required me adding a new parameter called `edge` which when set to False does not add halos at the edges (basically where you 'fake' the value using either zero, reflect, periodic

- I wanted to use Halo to implement a distributed version of https://pylops.readthedocs.io/en/stable/api/generated/pylops.signalprocessing.Sliding1D.html#pylops.signalprocessing.Sliding1D.. however I realized that in this case one would need to have halos that change from rank to rank.. I can sketch why, but more in general you think this is feasible?

astroC86:
> I am going to summarize my experience so far here and tomorrow I will push an updated version of `plot_halo` and a new `plot_halo1d` which uses an updated version of the actual implementation (will push it in a file called `Halo_new.py` so you can see what changes I made - so far they seem to break the 3d implementation though...):

Perfect,will have a look once you push!

> * I tried to check if the operator passes the dottest with this very simple example, but i got `AssertionError: Dot test failed, v^H(Opu)=61.67157626475711 - u^H(Op^Hv)=65.33767302699903`... not really sure why, but it seems that the adjoint is not correct (I trust the forward as I used in various examples and it always seem to make sense)

The reason being is the Halo adjoint is not a true "adjoint" , it simply removes the halo from the `local_submatirx` returning the core. I will try to think of a way to get the adjoint working as you expect it for the dottest so the adjoint is as you expect it. Thanks for the example by the way very handy to test things, really appreciate it!


> * I used Halo to implement a distributed FirstDerivative operator using only pylops FirstDerivative and MPIBlockDiag... I think this is a nice example of how one could use Halo instead of rewrite from scratch a distributed operator and use `add_ghost_cells`.. this however required me adding a new parameter called `edge` which when set to False does not add halos at the edges (basically where you 'fake' the value using either zero, reflect, periodic

Ok so if I understand correctly you want to have the ability to have no halos long the edges, is that correct? 

> * I wanted to use Halo to implement a distributed version of https://pylops.readthedocs.io/en/stable/api/generated/pylops.signalprocessing.Sliding1D.html#pylops.signalprocessing.Sliding1D.. however I realized that in this case one would need to have halos that change from rank to rank.. I can sketch why, but more in general you think this is feasible?

I will try to understand the `Sliding1D` operator before I give you a concrete, but I think it could be possible

pylops_mpi/basicoperators/Halo.py
```diff
@@ -0,0 +1,192 @@
...
+            arr[tuple(rcv_s)] = rcv
+        # send after
+        if after and plus_nbr != MPI.PROC_NULL:
+            snd_s = slicer.copy()
+            snd_s[axis] = slice(-2 * after, -after)
+            rcv_s = slicer.copy()
+            rcv_s[axis] = slice(-after, None)
+            snd = arr[tuple(snd_s)].copy()
+            rcv = ncp.empty_like(snd)
+            self.cart_comm.Sendrecv(snd, dest=plus_nbr, recvbuf=rcv, source=plus_nbr)
```
astroC86: use send or recv or port logic to nccl

pylops_mpi/utils/_common.py
```diff
@@ -1,58 +1,101 @@
...
+    )
+
+    ndev = len(send_buf_shapes)
+    recv_buf = ncp.zeros(ndev * send_buf.size, dtype=send_buf.dtype)
+
+    return send_buf, recv_buf
+
+
+def _unroll_allgather_recv(recv_buf, padded_send_buf_shape, send_buf_shapes, displs=None) -> list:
+    r"""Unrolll recv_buf after Buffered Allgather (MPI and NCCL)
```
mrava87: Unrolll -> Unroll

pylops_mpi/utils/_common.py
mrava87, line 1: I am a bit confused, the `_unroll_allgather_recv` seems very different from that of main (https://github.com/PyLops/pylops-mpi/blob/main/pylops_mpi/utils/_common.py#L8)

This seems more like what we have in v0.4.0 (https://github.com/PyLops/pylops-mpi/blob/v0.4.0/pylops_mpi/utils/_common.py) but not what we have in main... seems like you force overwrite some latest changes?

Please take a careful look.

pylops_mpi/basicoperators/VStack.py
```diff
@@ -183,7 +183,10 @@ def __init__(self, ops: Sequence[MPILinearOperator],
         self.ops = ops
         if len(set(op.shape[1] for op in ops)) > 1:
             raise ValueError("Operators have different number of columns")
-        shape = (int(sum(op.shape[0] for op in ops)), ops[0].shape[1])
+        shape = (
```
mrava87: Are you sure this is needed... I wonder if the numpy deprecation you saw (see from your commit message) was not the same that we encountered some time back and fixed here https://github.com/PyLops/pylops-mpi/pull/189/changes#diff-9db14bd1e016ff93cf721f37da3b6a9e641cc25f85f8ad1a9f9aebdb8197ba3f?

pylops_mpi/basicoperators/BlockDiag.py
```diff
@@ -170,8 +170,10 @@ def __init__(self, ops: Sequence[MPILinearOperator], base_comm: MPI.Comm = MPI.C
                  dtype: Optional[DTypeLike] = None):
         self.ops = ops
         dtype = _get_dtype(self.ops) if dtype is None else np.dtype(dtype)
-        shape = (int(sum(op.shape[0] for op in ops)),
-                 int(sum(op.shape[1] for op in ops)))
+        shape = (
```
mrava87: Same comment I left to VStack changes

pylops_mpi/StackedLinearOperator.py
```diff
@@ -103,7 +105,9 @@ def rmatvec(self, x: Union[DistributedArray, StackedDistributedArray]) -> Union[
 
         M, N = self.shape
         if isinstance(x, StackedDistributedArray):
-            stacked_shape = (np.sum([a.global_shape for a in x.distarrays]), )
+            stacked_shape = (
```
mrava87: Ok but this one we did not change... maybe it is a new warning.

I'll wait to hear from you and pinging also @rohanbabbar04 that worked on this PR (https://github.com/PyLops/pylops-mpi/pull/189/changes#diff-9db14bd1e016ff93cf721f37da3b6a9e641cc25f85f8ad1a9f9aebdb8197ba3f?)

pylops_mpi/basicoperators/Halo.py
```diff
@@ -0,0 +1,307 @@
...
+            int(np.prod(self.local_extent)),
+        )
+
+        self.shape = (
+            int(np.sum(self._local_extent_sizes)),
+            int(np.prod(self.global_dims)),
+        )
+        super().__init__(shape=self.shape, dtype=np.dtype(dtype), base_comm=comm)
+
+    def _parse_halo(self, h):
```
mrava87: Add small docstring explaining what this method does (can simply refer to the Notes, so say something like convert h from scalar to tuple of 2*ndim elements with halos along each axis)...

pylops_mpi/basicoperators/Halo.py
```diff
@@ -0,0 +1,307 @@
...
+    block, typically via :class:`pylops_mpi.basicoperators.MPIBlockDiag`.
+
+    In the adjoint mode, we perform the reverse where we extract the original
+    local domain by removing the ghost cells and returning the local core.
+    """
+
+    def __init__(
+        self,
+        dims: tuple,
+        halo,
```
mrava87: ideally try to always type annotate... we do still have places where we don't and we should improve them, but at least for new code it's good to keep the standard high 😄

pylops_mpi/basicoperators/Halo.py
```diff
@@ -0,0 +1,307 @@
...
+            halo = h * (2 * self.ndim)
+        elif len(h) == self.ndim:
+            halo = sum(tuple([(d, d) for d in h]), ())
+        elif len(h) == 2 * self.ndim:
+            halo = h
+        else:
+            raise ValueError(f"Invalid halo length {len(h)} for ndim={self.ndim}")
+        return halo
+
+    def _build_topo(self):
```
mrava87: Same as above, short description of what this method does

pylops_mpi/basicoperators/Halo.py
```diff
@@ -0,0 +1,307 @@
...
+            reorder=True,
+        )
+        neigh = {}
+        for ax in range(self.ndim):
+            before, after = cart_comm.Shift(ax, 1)
+            neigh[("-", ax)] = before
+            neigh[("+", ax)] = after
+        return cart_comm, neigh
+
+    def _calc_local_dims(self):
```
mrava87: Same

tests/test_halo.py
mrava87, line 1: There are not enough tests for the amount of features you implemented.

For example, we need to:

- test all 3 different combinations of `halo`: scalar, tuple with ndim and tuple with 2*ndim elements;
- 1d/2d/3d dims. I would ideally split that into 3 test functions
- test halo in isolation checking it produces an output with expected local shapes vs test halo combined with another operator (as you already do)
- test that a wrong grid_shape is catched (use the with pytest.raises.. type of test)

pylops_mpi/basicoperators/Halo.py
```diff
@@ -0,0 +1,307 @@
...
+
+import numpy as np
+from mpi4py import MPI
+from pylops.utils.backend import get_module
+
+from pylops_mpi import DistributedArray, MPILinearOperator, Partition
+from pylops_mpi.Distributed import DistributedMixIn
+
+
+def halo_block_split(global_shape: tuple, comm, grid_shape: tuple = None) -> tuple:
```
mrava87: add type hint to `comm`

mrava87 (changes requested):
@astroC86 great work!

I reviewed the PR and I think we are very close for this to be ready to be merged 😄 

I left a few small comments about doing some small improvements to type annotations/docstrings and a couple about changes in places that don't seem to match with what i see in the current main branch (probably you did work on those files as others PRs come in, so I just want to make sure we are not undoing any other concurrent work by mistake)

pylops_mpi/StackedLinearOperator.py
rohanbabbar04: Hmm.. 
It is not giving any issues/warnings. I think both will give identical results because our `global_shape` is 1-D. 

I would suggest we use `sum` like we have done elsewhere as well.

```
stacked_shape = (sum([a.global_shape[0] for a in x.distarrays]), )
```

mrava87: Yeah i tend to agree, unless you identify something new that we didn't... ideally this would not be the best PR to fix that anyways

pylops_mpi/utils/_common.py
mrava87: @astroC86 _unroll_allgather_recv still seems to be very different from the main (https://github.com/PyLops/pylops-mpi/blob/main/pylops_mpi/utils/_common.py#L8), why?

mrava87 (approved):
@astroC86 I think this is ready to be merged, great job 🚀

Issue #172, opened 2025-10-30
Bug: StackedDistributedArray norm returns error for arrays with inconsistent `shape`.


mrava87:
https://github.com/PyLops/pylops-mpi/blob/5362ed27c989d8c90b62de689d6c017dfe5e40c2/pylops_mpi/DistributedArray.py#L980

does return the following error `TypeError: Implicit conversion to a NumPy array is not allowed. Please use `.get()` to construct a NumPy array explicitly.`

if the different norms being put in the `norms` array are not-consistent in terms of shapes. This can be explained more easily with an example:

```
cp.linalg.norm(cp.ones(10)), cp.ones(1)
> (array(3.16227766), array([1.]))
cp.linalg.norm(cp.ones(10)).shape, cp.ones(1).shape
> ((), (1,))
```

basically one is a zero element array and the other is a 1 element one...

Now, this situation seems to arise only in the NCCL scenario if the distributed arrays where norm is computed on have dtype of the form `float32` vs `np.float32`.

This small change seems to fix the problem `norms = ncp.array([distarray.norm(ord).item() for distarray in self.distarrays])` but I would like this to be investigated a bit more broadly to make sure there is no other method in `StackedDistributedArray` where something similar happens.

PR #173, opened 2025-11-03
bug: pass base_comm_nccl to __neg__ 

mrava87:
Fixes https://github.com/PyLops/pylops-mpi/issues/172

Moreover, python 3.10 is removed from the CI (leaving 3.11,3.12, and 3.13)


mrava87:
I just felt the CI was taking way too long and I don’t know anymore that still uses Python 3.10… since we may soon want to start adding Python 3.14, I think having 5 is really too many 😂

(I also thought/felt the MatrixMult failures were always on 3.10 but turns out it is not…)

PR #192, opened 2026-04-03
ISTA/FISTA implementation in cls_sparsity

### LINKED ISSUE #97: FISTA solver
Add `cls_sparsity` and `sparsity` submodule to `optimization` starting with the implementation of `ISTA` and `FISTA`. 

This solvers could be easily tested on a slightly modified version of our tutorial in `poststack.py` following this notebook: https://github.com/PyLops/pylops_notebooks/blob/master/developement/SeismicInversion-reflectivity.ipynb

rohanbabbar04:
Closes #97 
- [x] ISTA added to cls_sparsity
- [x] `power_iteration` to calculate maxeig.
- [x] Tests for power_iteration and ISTA.
- [x] Reflectivity tutorial(s) - base, cupy and nccl
- [x] FISTA
- [x] Add support for Stacked Operators.
- [x] Fix `Partition.BROADCAST` issue

### DIFF
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -8,7 +8,7 @@ dependencies:
   - numpy>=1.15.0
   - scipy>=1.8.0
   - mpi4py
-  - pylops
+  - pylops>=2.6.0
   - matplotlib
   - ipython
   - pytest
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -8,7 +8,7 @@ dependencies:
   - numpy>=1.15.0
   - scipy>=1.8.0
   - mpi4py
-  - pylops>=2.6.0
+  - pylops>=2.0.0
   - matplotlib
   - ipython
   - pytest
--- a/pylops_mpi/DistributedArray.py
+++ b/pylops_mpi/DistributedArray.py
@@ -205,12 +205,20 @@ class DistributedArray(DistributedMixIn):
             the specified index positions.
         """
         if self.partition is Partition.BROADCAST:
+            ncp = get_module(self.engine)
             view = self.local_array[index]
+            shape = getattr(view, "shape", ())
+            buf = ncp.empty(shape, dtype=self.dtype)
             if self.rank == 0:
-                view[...] = value
-            view = self._bcast(self.base_comm, self.base_comm_nccl,
-                               view, root=0, engine=self.engine)
-            self.local_array[index] = view
+                buf[...] = value
+            buf = self._bcast(
+                self.base_comm,
+                self.base_comm_nccl,
+                buf,
+                root=0,
+                engine=self.engine
+            )
+            self.local_array[index] = buf
         else:
             self.local_array[index] = value
 
@@ -644,7 +652,7 @@ class DistributedArray(DistributedMixIn):
             ProductArray[:] = self.local_array * dist_array
         return ProductArray
 
-    def dot(self, dist_array):
+    def dot(self, dist_array, vdot=False):
         """Distributed Dot Product
         """
         self._check_partition_shape(dist_array)
@@ -656,8 +664,12 @@ class DistributedArray(DistributedMixIn):
         y = DistributedArray.to_dist(x=dist_array.local_array, base_comm=self.base_comm, base_comm_nccl=self.base_comm_nccl) \
             if self.partition in [Partition.BROADCAST, Partition.UNSAFE_BROADCAST] else dist_array
         # Flatten the local arrays and calculate dot product
+        if vdot:
+            dot_func = ncp.vdot
+        else:
+            dot_func = ncp.dot
         return self._allreduce_subcomm(self.sub_comm, self.base_comm_nccl,
-                                       ncp.dot(x.local_array.flatten(), y.local_array.flatten()),
+                                       dot_func(x.local_array.flatten(), y.local_array.flatten()),
                                        engine=self.engine)
 
     def _compute_vector_norm(self, local_array: NDArray,
@@ -838,6 +850,15 @@ class DistributedArray(DistributedMixIn):
         x = local_array.copy()
         arr[:] = x
         return arr
+
+    def empty_like(self):
+        """Creates an empty like DistributedArray with uninitialized values
+        """
+        dist = DistributedArray(global_shape=self.global_shape, base_comm=self.base_comm,
+                                base_comm_nccl=self.base_comm_nccl, partition=self.partition,
+                                axis=self.axis, local_shapes=self.local_shapes, mask=self.mask,
+                                engine=self.engine, dtype=self.dtype)
+        return dist
 
     def add_ghost_cells(self, cells_front: Optional[int] = None,
                         cells_back: Optional[int] = None):
@@ -1038,13 +1059,13 @@ class StackedDistributedArray:
                 ProductArray[iarr][:] = (self[iarr] * stacked_array)[:]
         return ProductArray
 
-    def dot(self, stacked_array):
+    def dot(self, stacked_array, vdot=False):
         """Dot Product of Stacked Distributed Arrays
         """
         self._check_stacked_size(stacked_array)
         dotprod = 0.
         for iarr in range(self.narrays):
-            dotprod += self[iarr].dot(stacked_array[iarr])
+            dotprod += self[iarr].dot(stacked_array[iarr], vdot=vdot)
         return dotprod
 
     def norm(self, ord: Optional[int] = None):
@@ -1084,6 +1105,19 @@ class StackedDistributedArray:
         """
         arr = StackedDistributedArray([distarray.copy() for distarray in self.distarrays])
         return arr
+
+    def empty_like(self):
+        """Creates an empty like StackedDistributedArray with uninitialized values
+        """
+        dists = []
+        for iarr in range(self.narrays):
+            distarray = self.distarrays[iarr]
+            dist = DistributedArray(global_shape=distarray.global_shape, base_comm=distarray.base_comm,
+                                    base_comm_nccl=distarray.base_comm_nccl, partition=distarray.partition,
+                                    axis=distarray.axis, local_shapes=distarray.local_shapes, mask=distarray.mask,
+                                    engine=distarray.engine, dtype=distarray.dtype)
+            dists.append(dist)
+        return StackedDistributedArray(distarrays=dists)
 
     def __repr__(self):
         repr_dist = "\n".join([distarray.__repr__() for distarray in self.distarrays])
--- a/pylops_mpi/DistributedArray.py
+++ b/pylops_mpi/DistributedArray.py
@@ -207,8 +207,7 @@ class DistributedArray(DistributedMixIn):
         if self.partition is Partition.BROADCAST:
             ncp = get_module(self.engine)
             view = self.local_array[index]
-            shape = getattr(view, "shape", ())
-            buf = ncp.empty(shape, dtype=self.dtype)
+            buf = ncp.empty(view.shape, dtype=self.dtype)
             if self.rank == 0:
                 buf[...] = value
             buf = self._bcast(
@@ -652,8 +651,25 @@ class DistributedArray(DistributedMixIn):
             ProductArray[:] = self.local_array * dist_array
         return ProductArray
 
-    def dot(self, dist_array, vdot=False):
-        """Distributed Dot Product
+    def dot(self, dist_array, vdot: bool = False):
+        """
+        Compute the distributed dot product between this array and another
+        distributed array.
+
+        Parameters
+        ----------
+        dist_array : :obj:`pylops_mpi.DistributedArray`
+            The distributed array with which to compute the dot product.
+            It must have a compatible shape and partitioning.
+        vdot : bool, optional, Defaults to `False`
+            If True, compute the complex conjugate dot product (like numpy.vdot),
+            where the first argument is conjugated. If False, compute the standard
+            dot product.
+
+        Returns
+        -------
+        result : float
+            The result of the dot product across all ranks. This is reduced across all processes.
         """
         self._check_partition_shape(dist_array)
         self._check_mask(dist_array)
@@ -664,10 +680,7 @@ class DistributedArray(DistributedMixIn):
         y = DistributedArray.to_dist(x=dist_array.local_array, base_comm=self.base_comm, base_comm_nccl=self.base_comm_nccl) \
             if self.partition in [Partition.BROADCAST, Partition.UNSAFE_BROADCAST] else dist_array
         # Flatten the local arrays and calculate dot product
-        if vdot:
-            dot_func = ncp.vdot
-        else:
-            dot_func = ncp.dot
+        dot_func = ncp.vdot if vdot else ncp.dot
         return self._allreduce_subcomm(self.sub_comm, self.base_comm_nccl,
                                        dot_func(x.local_array.flatten(), y.local_array.flatten()),
                                        engine=self.engine)
@@ -1059,8 +1072,25 @@ class StackedDistributedArray:
                 ProductArray[iarr][:] = (self[iarr] * stacked_array)[:]
         return ProductArray
 
-    def dot(self, stacked_array, vdot=False):
-        """Dot Product of Stacked Distributed Arrays
+    def dot(self, stacked_array, vdot: bool = False):
+        """
+        Compute the distributed dot product between this array and another
+        distributed array.
+
+        Parameters
+        ----------
+        stacked_array : :obj:`pylops_mpi.StackedDistributedArray`
+            The distributed array with which to compute the dot product.
+            It must have a compatible shape and partitioning.
+        vdot : bool, optional, Defaults to `False`
+            If True, compute the complex conjugate dot product (like numpy.vdot),
+            where the first argument is conjugated. If False, compute the standard
+            dot product.
+
+        Returns
+        -------
+        result : float
+            The result of the dot product across all ranks. This is reduced across all processes.
         """
         self._check_stacked_size(stacked_array)
         dotprod = 0.
--- a/pylops_mpi/optimization/cls_sparsity.py
+++ b/pylops_mpi/optimization/cls_sparsity.py
@@ -6,9 +6,9 @@ from typing import Optional, Dict, Any, Callable
 from pylops.optimization.basesolver import Solver
 from pylops.optimization.callback import _callback_stop
 from pylops.utils import get_array_module, get_module_name, get_real_dtype
-from pylops.utils.typing import NDArray, Union, Tuple
+from pylops.utils.typing import NDArray, Tuple
 
-from pylops_mpi.DistributedArray import DistributedArray, StackedDistributedArray
+from pylops_mpi.DistributedArray import DistributedArray
 from pylops_mpi.LinearOperator import MPILinearOperator
 from pylops_mpi.optimization.eigs import power_iteration
 
@@ -225,13 +225,11 @@ class ISTA(Solver):
 
     def _print_step(
             self,
-            x: Union[DistributedArray, StackedDistributedArray],
+            x: DistributedArray,
             costdata: float,
             costreg: float,
             xupdate: float,
     ) -> None:
-        if isinstance(x, DistributedArray):
-            x = x.distarrays[0]
         strx = (
             f"  {x[0]:1.2e}   " if np.iscomplexobj(x.local_array) else f"     {x[0]:11.4e}        "
         )
@@ -251,8 +249,8 @@ class ISTA(Solver):
 
     def setup(
             self,
-            y: Union[DistributedArray, StackedDistributedArray],
-            x0: Union[DistributedArray, StackedDistributedArray],
+            y: DistributedArray,
+            x0: DistributedArray,
             niter: Optional[int] = None,
             SOp: Optional[MPILinearOperator] = None,
             eps: float = 0.1,
@@ -265,14 +263,14 @@ class ISTA(Solver):
             monitorres: bool = False,
             preallocate: bool = False,
             show: bool = False,
-    ) -> Union[DistributedArray, StackedDistributedArray]:
+    ) -> DistributedArray:
         r"""Setup solver
 
         Parameters
         ----------
-        y : :obj:`pylops_mpi.DistributedArray` or :obj:`pylops_mpi.StackedDistributedArray`
+        y : :obj:`pylops_mpi.DistributedArray`
             Data of size (N, )
-        x0 : :obj:`pylops_mpi.DistributedArray` or :obj:`pylops_mpi.StackedDistributedArray`
+        x0 : :obj:`pylops_mpi.DistributedArray``
             Initial guess  of size (M, ).
         niter : :obj:`int`
             Number of iterations
@@ -309,7 +307,7 @@ class ISTA(Solver):
 
         Returns
         -------
-        x : :obj:`pylops_mpi.DistributedArray` or :obj:`pylops_mpi.StackedDistributedArray`
+        x : :obj:`pylops_mpi.DistributedArray`
             Initial guess of size (N,).
 
         """
@@ -397,20 +395,20 @@ class ISTA(Solver):
             self._print_setup()
         return x
 
-    def step(self, x: Union[DistributedArray, StackedDistributedArray], show: bool = False) -> (
-            Tuple)[Union[DistributedArray, StackedDistributedArray], float]:
+    def step(self, x: DistributedArray, show: bool = False) -> (
+            Tuple)[DistributedArray, float]:
         r"""Run one step of solver
 
         Parameters
         ----------
-        x : :obj:`pylops_mpi.DistributedArray` or :obj:`pylops_mpi.StackedDistributedArray`
+        x : :obj:`pylops_mpi.DistributedArray`
             Current model vector to be updated by a step of ISTA
         show : :obj:`bool`, optional
             Display iteration log
 
         Returns
         -------
-        x : :obj:`pylops_mpi.DistributedArray` or :obj:`pylops_mpi.StackedDistributedArray`
+        x : :obj:`pylops_mpi.DistributedArray`
             Updated model vector
         xupdate : :obj:`float`
             Norm of the update
@@ -488,16 +486,16 @@ class ISTA(Solver):
 
     def run(
             self,
-            x: Union[DistributedArray, StackedDistributedArray],
+            x: DistributedArray,
             niter: Optional[int] = None,
             show: bool = False,
             itershow: Tuple[int, int, int] = (10, 10, 10),
-    ) -> Union[DistributedArray, StackedDistributedArray]:
+    ) -> DistributedArray:
         r"""Run solver
 
         Parameters
         ----------
-        x : :obj:`pylops_mpi.DistributedArray` or :obj:`pylops_mpi.StackedDistributedArray`
+        x : :obj:`pylops_mpi.DistributedArray`
             Current model vector to be updated by multiple steps of ISTA
         niter : :obj:`int`, optional
             Number of iterations. Can be set to ``None`` if already
@@ -511,7 +509,7 @@ class ISTA(Solver):
 
         Returns
         -------
-        x : :obj:`pylops_mpi.DistributedArray` or :obj:`pylops_mpi.StackedDistributedArray
+        x : :obj:`pylops_mpi.DistributedArray`
             Estimated model of size (M, ).
 
         """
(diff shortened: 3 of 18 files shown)


rohanbabbar04:
@mrava87, have pushed in the commits which implement ISTA using mpi(numpy and cupy). 
Some things which need to be implemented:
1) Get it working for nccl (need to see how much this needs to change).
2) Currently, the function does not support thresh kind of hard_percentile, soft_percentile and half_percentile(need to see how it can be implemented in a distributed manner, any suggestions on how we can calculate percentile using mpi).

mrava87:
@rohanbabbar04 nice, I'll try to have a look in coming days 😄 

Regarding your questions:

1. What type of issues are you facing with NCCL? At first glance I would expect ISTA/FISTA not to be any different communication-wise from the CG/CGLS solvers we have already, but I may have overlooked something...
2. These are fairly exotic choices of threshold, not the most common ones for ISTA solvers. I would say, let's just not bother about them for now.. I can see that doing distributed sorting isn't the most straightforward and cheap of operations and percentiles really require an array to be sorted to be able to identify the percentile of the array.
3. I saw you had a comment about tests, but I guess you took it away. For tests, like all other tests in pylops-mpi, I think it is important to have runs against the serial solver and verify they produce the same result. We could use the problem setup of pylops and keep the test similar but the check is what matters the most 😄

rohanbabbar04:
Thanks for a quick reply.

1) I thought so as well. I think I may have missed something for nccl which is why it was giving me an error when I tried to push the test to the CI. Let me push the test and we can figure out why is that happening.
2) Great, let's keep the percentile out for now.
3) Yeah, I actually realised that myself that's why I removed the message. I will add the other types of ISTA tests which we have in pylops 🙂.

rohanbabbar04:
@mrava87, most of the implementation is in place.
1) Nccl is now working, made some minor changes(np to ncp) after I figured out why it was not working.
2) The tests are similar to what we have in pylops.
3) Only thing which needs to be added is the support for Stacked Operators. It would mean extending the code and functions such as power_iteration, thresholds to support that. I will work on that once you are able to go though the code.
Thanks 🙂

pylops_mpi/optimization/eigs.py
```diff
@@ -0,0 +1,74 @@
+from mpi4py import MPI
+from pylops.utils.backend import get_module
+from pylops_mpi.DistributedArray import DistributedArray, NcclCommunicatorType
+from pylops_mpi.LinearOperator import MPILinearOperator
+from typing import Tuple, Optional
```
mrava87: from typing should go at the top as it is part of the standard python library

pylops_mpi/optimization/eigs.py
```diff
@@ -0,0 +1,74 @@
...
+
+    Parameters
+    ----------
+    Op : :obj:`pylops_mpi.MPILinearOperator`
+        Square operator
+    niter : :obj:`int`, optional
+        Number of iterations
+    tol : :obj:`float`, optional
+        Update tolerance
+    base_comm : :obj:`MPI.Comm`
```
mrava87: add `, optional`

Same for the next one

pylops_mpi/optimization/eigs.py
```diff
@@ -0,0 +1,74 @@
...
+    b_k : :obj:`pylops_mpi.DistributedArray`
+        Largest eigenvector
+    iiter : :obj:`int`
+        Effective number of iterations
+
+    """
+
+    ncp = get_module(backend)
+    cmpx = 1j if ncp.issubdtype(ncp.dtype(dtype), ncp.complexfloating) else 0
+    b_k = DistributedArray(global_shape=Op.shape[1], dtype=dtype, engine=backend,
```
mrava87: Mmh you do not specify the local shape and  partition here? I assume for this to work the input must have the partition and local shape that the operator expects... I guess we need to pass them...

pylops_mpi/optimization/eigs.py
```diff
@@ -0,0 +1,74 @@
...
+                           base_comm=base_comm, base_comm_nccl=base_comm_nccl)
+    b_k[:] = (
+        ncp.random.rand(b_k.local_shape[0]).astype(dtype)
+        + cmpx * ncp.random.rand(b_k.local_shape[0]).astype(dtype)
+    )
+    b_k[:] /= b_k.norm()
+    maxeig_old = 0.0
+    for iiter in range(niter):
+        b1_k = Op.matvec(b_k)
+        # Calculate vdot for the global array
```
mrava87: I suggest modifying `dot` in DistributedArray adding a parameter vdot=False that when true uses vdot instead of dot, so we can inherit the dot operation instead of having to add local+comm here...

mrava87: Would be good to reinstate the same comments we have in pylops 😄 comment don't cost anything and they help other people follow a code (20 lines of pure code are never easy to follow unless one tries to read line by line and figure out what is happening...)

pylops_mpi/optimization/eigs.py
```diff
@@ -0,0 +1,74 @@
...
+        local_dot = ncp.vdot(b_k.local_array, b1_k.local_array)
+        maxeig = b_k._allreduce(b_k.base_comm, b_k.base_comm_nccl, local_dot, engine=backend).item()
+        b_k[:] = b1_k.local_array / b1_k.norm()
+        if ncp.abs(maxeig - maxeig_old) < tol * ncp.abs(maxeig):
```
mrava87: mmh why not `< maxeig_old` like in pylops? it is probably going to be the same / very similar as one should not expect a huge change from iteration to iteration, but I would rather keep consistency unless there is a good reason not to...

pylops_mpi/optimization/cls_sparsity.py
```diff
@@ -0,0 +1,638 @@
...
+from pylops.utils.typing import NDArray, Tuple
+
+from pylops_mpi.DistributedArray import DistributedArray
+from pylops_mpi.LinearOperator import MPILinearOperator
+from pylops_mpi.optimization.eigs import power_iteration
+
+logger = logging.getLogger(__name__)
+
+
+def _hardthreshold(x: DistributedArray, thresh: float) -> DistributedArray:
```
mrava87: At first, I would have expected that the thresholding methods did not need reimplemented as they act element-wise so if you have a numpy array vs distributed array things should not change.

What about doing something like this in the solver where these methods are called:

```
x = x_unthesh_or_SOpx_unthesh.copy()
x.local_array[:] = self.threshf(x.local_array, self.thresh)
```

where the `self.threshf` points to the original methods in PyLops instead of the re-implemented ones.

The reason I am very keen for this is that the less redundant code the better; if one day we were to make a change in PyLops we may forget to make the same in pylops-mpi, so where there is no need to duplicate better not to 😄

pylops_mpi/optimization/cls_sparsity.py
```diff
@@ -0,0 +1,638 @@
...
+    # Set elements to 0 where their magnitude is less than or equal to sqrt(2 * thresh) at each rank
+    local_out[ncp.abs(local) <= ncp.sqrt(2 * thresh)] = 0.0
+
+    dist = DistributedArray(global_shape=x.global_shape, partition=x.partition, base_comm=x.base_comm,
+                            base_comm_nccl=x.base_comm_nccl, dtype=x.dtype, engine=x.engine)
+    dist[:] = local_out
+    return dist
+
+
+def _softthreshold(x: DistributedArray, thresh: float) -> DistributedArray:
```
mrava87: Same as above... for now I do not reviews these 2 functions, if you don't think my proposal of getting rid of them makes sense I will review them.

pylops_mpi/optimization/cls_sparsity.py
mrava87: One more idea that just came to my mind is that by doing this we could more easily be able to also enable stacked array.. if we create an internal method `_apply_thresh` or something like that we could put there the logic... eg

```
def _apply_thresh(self, x, threshf, thresh):
        if isinstance(x, DistributedArray):
                  x.local_array[:] = threshf(x.local_array, thresh)
        else:
            for iarr in range(x.narrays):
                  x[iarr].local_array[:] += threshf(x[iarr].local_array, thresh)
        return x
```

and then

```
x = self._apply_thresh(x_unthesh_or_SOpx_unthesh.copy(), self.threshf)
```

pylops_mpi/optimization/cls_sparsity.py
```diff
@@ -0,0 +1,638 @@
...
+    Op : :obj:`pylops_mpi.MPILinearOperator`
+        Operator to invert
+
+    Attributes
+    ----------
+    ncp : :obj:`module`
+        Array module used by the solver (obtained via
+        :func:`pylops.utils.backend.get_array_module`)
+        ). Available only after ``setup`` is called.
+    isjax : :obj:`bool`
```
mrava87: Remove this, I don' think it applies here

pylops_mpi/optimization/cls_sparsity.py
```diff
@@ -0,0 +1,638 @@
...
+        If ``threshkind`` is different from hard, soft, half
+    """
+
+    def _print_setup(self) -> None:
+        self._print_solver(f" ({self.threshkind} thresholding)")
+        if self.niter is not None:
+            strpar = f"eps = {self.eps:10e}\ttol = {self.tol:10e}\tniter = {self.niter}"
+        else:
+            strpar = f"eps = {self.eps:10e}\ttol = {self.tol:10e}"
+        if self.perc is None:
```
mrava87: I thought we were not going to support percentiles?

pylops_mpi/optimization/cls_sparsity.py
```diff
@@ -0,0 +1,638 @@
...
+            y: DistributedArray,
+            x0: DistributedArray,
+            niter: Optional[int] = None,
+            SOp: Optional[MPILinearOperator] = None,
+            eps: float = 0.1,
+            alpha: Optional[float] = None,
+            eigsdict: Optional[Dict[str, Any]] = None,
+            tol: float = 1e-10,
+            threshkind: str = "soft",
+            perc: Optional[float] = None,
```
mrava87: Remove this if not supported?

mrava87: Remove?

pylops_mpi/optimization/cls_sparsity.py
```diff
@@ -0,0 +1,638 @@
...
+            convergence criterion will not be checked internally.
+        eigsdict : :obj:`dict`, optional
+            Dictionary of parameters to be passed to power_iteration` method
+            when computing the maximum eigenvalue
+        tol : :obj:`float`, optional
+            Tolerance. Stop iterations if difference between inverted model
+            at subsequent iterations is smaller than ``tol``
+        threshkind : :obj:`str`, optional
+            Kind of thresholding ('hard', 'soft', 'half' - 'soft' used as default)
+        perc : :obj:`float`, optional
```
mrava87: Remove

pylops_mpi/optimization/cls_sparsity.py
```diff
@@ -0,0 +1,638 @@
...
+        """
+
+        self.y = y
+        self.SOp = SOp
+        self.niter = niter
+        self.eps = eps
+        self.eigsdict = {} if eigsdict is None else eigsdict
+        self.tol = tol
+        self.threshkind = threshkind
+        self.perc = perc
```
mrava87: Remove

pylops_mpi/optimization/cls_sparsity.py
```diff
@@ -0,0 +1,638 @@
...
+            self.SOpmatvec = self.SOp.matvec
+            self.SOprmatvec = self.SOp.rmatvec
+
+        if threshkind not in [
+            "hard",
+            "soft",
+            "half",
+        ]:
+            raise ValueError(
+                "threshkind must be hard, soft, half,"
```
mrava87: add `, got {threshkind}"`

pylops_mpi/optimization/cls_sparsity.py
```diff
@@ -0,0 +1,638 @@
...
+
+        self.threshf: Callable[[DistributedArray, float], DistributedArray]
+        if threshkind == "soft":
+            self.threshf = _softthreshold
+        elif threshkind == "hard":
+            self.threshf = _hardthreshold
+        elif threshkind == "half":
+            self.threshf = _halfthreshold
+
+        if perc is None and decay is None:
```
mrava87: Remove perc

pylops_mpi/optimization/cls_sparsity.py
```diff
@@ -0,0 +1,638 @@
...
+            self.decay = self.ncp.ones(niter, dtype=get_real_dtype(self.Op.dtype))
+        if alpha is not None:
```
mrava87, line 239: Everywhere, reinstate the same comments as pylops solver

pylops_mpi/optimization/cls_sparsity.py
```diff
@@ -0,0 +1,638 @@
...
+            self,
+            show: bool = False,
+            unit: str = "B",
+    ) -> float:
+        pass
+
+    def setup(
+            self,
+            y: DistributedArray,
+            x0: DistributedArray,
```
mrava87: You took away the = None but then allow this to be None (see below)

mrava87: However I think that if alpha is not provided we can't get away with a user not providing x0 as we will not be able to find the correct local shape and partition for `power_iteration` (see comment to power iteration). If this is the case, we need to document this in the docstring and have a check that if x0=None and alpha=None we need to raise an error...

pylops_mpi/optimization/cls_sparsity.py
```diff
@@ -0,0 +1,638 @@
...
+            self.alpha = float(1.0 / maxeig)
+        self.thresh = eps * self.alpha * 0.5
+        if x0 is None:
+            x = DistributedArray(global_shape=self.Op.shape[1], dtype=self.Op.dtype, engine=y.engine,
+                                 base_comm=y.base_comm, base_comm_nccl=y.base_comm_nccl)
+            x[:] = 0
+        else:
+            x = x0.copy()
+        if self.preallocate:
+            self.res = DistributedArray(global_shape=y.global_shape, dtype=y.dtype, engine=y.engine,
```
mrava87: again, don't we need to also pass local_shapes and partition of y?

Same to all other distributed arrays below

pylops_mpi/optimization/cls_sparsity.py
```diff
@@ -0,0 +1,638 @@
...
+
+        """
+        if self.preallocate:
+            self.xold[:] = x[:]
+        else:
+            xold = x.copy()
+        if not self.preallocate:
+            res = self.y - self.Opmatvec(x)
+        else:
+            self.res = self.y - self.Opmatvec(x)
```
mrava87: I think we should just remove the preallocate option as we don't seem to be leveraging that here... in pylops this is a fairly new feature (in fact if you look at our CG/CGLS in pylops-mpi we don't have it but pylops  CG/CGLS now have it).

In fact, `self.res = self.y - self.Opmatvec(x)` does not do anything in place and I suspect the fact we initialized `self.res` above has added some overhead instead of benefit... in future we could consider supporting operations in place but we would need to have methods like np.add/np.subtract that take out...

pylops_mpi/optimization/cls_sparsity.py
```diff
@@ -0,0 +1,638 @@
...
+        self.tend = time.time()
+        self.telapsed = self.tend - self.tstart
+        self.cost = np.array(self.cost)
+        if show:
+            self._print_finalize()
+
+    def solve(
+            self,
+            y: DistributedArray,
+            x0: Optional[DistributedArray] = None,
```
mrava87: Similar to the comment in setup... I do not see this being optional here as we can't infer the correct local_shapes/partition directly from the operator or y, isn't it?

In fact, I looked at https://github.com/PyLops/pylops-mpi/blob/901b706a5dc0e5755f4a476b910dd3e5dd07a73b/pylops_mpi/optimization/cls_basic.py#L473 and there we don't allow it to be optional

pylops_mpi/optimization/sparsity.py
```diff
@@ -0,0 +1,140 @@
...
+    rtol : :obj:`float`, optional
+        Relative tolerance on total cost function wrt initial total cost
+        function. Stops the solver when the ratio of the current total cost function
+        to the initial total cost function is below this value.
+    rtol1 : :obj:`float`, optional
+        Relative tolerance on total cost function wrt to data. Stops the solver when
+        the ratio of the current total cost function to the data norm is below this value.
+    threshkind : :obj:`str`, optional
+        Kind of thresholding ('hard', 'soft', 'half' - 'soft' used as default)
+    perc : :obj:`float`, optional
```
mrava87: Remove

tests/test_eigs.py
```diff
@@ -0,0 +1,43 @@
...
+from pylops_mpi.basicoperators.BlockDiag import MPIBlockDiag
+from pylops_mpi.optimization.eigs import power_iteration
+
+par1 = {"n": 21, "imag": 0, "dtype": "float64"}  # square, real
+par2 = {"n": 21, "imag": 1j, "dtype": "complex128"}  # square, complex
+
+
+@pytest.mark.parametrize("par", [(par1), (par2)])
+def test_power_iteration(par):
+    """Max eigenvalue computation with power iteration method vs. scipy methods"""
```
mrava87: there is no scipy method in here?

tests_nccl/test_sparsity_nccl.py
```diff
@@ -0,0 +1,196 @@
+"""Test Sparsity solvers
+    Designed to run with n GPUs (with 1 MPI process per GPU)
+    $ mpiexec -n 10 pytest test_sparsity_nccl.py --with-mpi
+
+This file employs the same test sets as test_sparsity_nccl under NCCL environment
```
mrava87: test_sparsity_nccl  -> test_sparsity

mrava87 (changes requested):
@rohanbabbar04 good start!

I have left a number of comments, some are minor but a few may need some thinking. More specifically:

- do we really need to reimplement the thresholding methods (see more detailed comments inline)... also if you see one of my comments, this may lead us to a much easier extension to StackedArray.
- are we really benefitting from preallocate as of now... I do not think so and I think this may require a bit of work on the DistributedArray side to become really the counterpart of pylops preallocate mode.... so for now I would just suggest we don't support it and take it away
- callback_stop: if we add this we need to bump the minimum required pylops version to 2.6.0
- comments: no idea why you stripped out all the comments that were in the pylops solver making the solver much harder to follow as it is just code now... reinstate all of them plus add any specific comment in parts where specific changes / modifications are made to enable the solver to work with distributed arrays
- FISTA: given that ISTA and FISTA are very close and FISTA inherits pretty much all the methods but the step, it would be great if we could add it at the same time, I do not foresee any major roadblock here 😄
- tests: I think that if we had tests also with a VStack operators that requires a model to have partition.BROADCAST we would have probably got some failures due to the fact that the some of DistrubutedArray that we create in the solver do not get local_shapes/partition specified.. it would be good to have tests with both set of partitions to make sure the solver is generic enough and here it will also help fixing any current bug (if my thinking is correct).
- running the solver with show=True seems to give me an error on the print of `xupdate`.
- add ISTA/ista to documentation API section

In parallel, I got a tutorial in place using ISTA as promised 😉 and this is where I discovered the issue with show=True...

rohanbabbar04:
Ah! I see that my docstring agent makes a lot of mistakes. I have turned it off :sweat_smile: .

Also, Thank you for the tutorial, it looks great 👍 

- Yes, I just realized we can do that and even better to implement an `apply_thresh` method for both `Distributed` and `StackedDistributedArray`.
- I was also a little bit confused about the benefit of `preallocate` as well, I think its better we remove it for now like you suggested.
- What would you suggest do we keep the callback_stop and bump the pylops to 2.6?
- Restored the comments.
- FISTA is now in, and yeah not much extra work. I agree better to include it over here only.
- Have added tests for HStack and VStack
- It should now work with `show=True`.
- Added ISTA and FISTA to API docs.

Some other changes
- Added FISTA, fista along with its tests.
- Updated `power_iteration` to include a initial guess `b_k`, which will be provided to the method to compute `Op.matvec(b_k)`.
- Also, in one of your comments I saw your suggestion about x0. I think its better to get the x0 from user like we have done for cgls.
- Have added `vdot` as a parameter in dot function.
- Have added `empty_like`, similar to what we have in numpy.
- One more change I made was to the broadcast.

Eerlier when I did something like this 
```python
x = DistributedArray(global_shape=size * par['nx'], dtype=par['dtype'], engine=backend)
x[:] = np.zeros(par["nx"], dtype=par['dtype']) + par["imag"] * np.zeros(par["nx"], dtype=par['dtype'])
x[par["nx"] // 2] = 1.0 + par["imag"] * 1.0
^^
# view[...] = value gives error    
````
I think this is because `self.local_array[index]` would be an `np.float` and it cannot assign the value(looks like it is due to the new np version). I made a change to create a buffer - `buf = ncp.empty(shape, dtype=self.dtype)` before we assign the value. 
Do let me know if you agree with this :slightly_smiling_face: ?

File: pylops_mpi/optimization/eigs.py (hunk 4 above)
Comment (rohanbabbar04) [reply pylops_mpi/optimization/eigs.py]: Done

Comment (rohanbabbar04) [reply pylops_mpi/optimization/eigs.py]: Sounds good

File: pylops_mpi/optimization/cls_sparsity.py (hunk 6 above)
Comment (rohanbabbar04) [reply pylops_mpi/optimization/cls_sparsity.py]: I have implemented  the `apply_thresh` method.

File: pylops_mpi/optimization/cls_sparsity.py (hunk 16 above)
Comment (rohanbabbar04) [reply pylops_mpi/optimization/cls_sparsity.py]: I would say we make x0 mandatory like we have in cgls.

File: pylops_mpi/optimization/cls_sparsity.py (hunk 9 above)
Comment (rohanbabbar04) [reply pylops_mpi/optimization/cls_sparsity.py]: Fixed

File: pylops_mpi/optimization/cls_sparsity.py (hunk 16 above)
Comment (rohanbabbar04) [reply pylops_mpi/optimization/cls_sparsity.py]: I have made a change to the `power_iteration` function, as it would be better that someone provides the initial guess `b_k` and then we can calculate the maxeig.

File: pylops_mpi/DistributedArray.py [hunk 23]
```diff
@@ -205,12 +205,20 @@ def __setitem__(self, index, value):
             the specified index positions.
         """
         if self.partition is Partition.BROADCAST:
+            ncp = get_module(self.engine)
             view = self.local_array[index]
+            shape = getattr(view, "shape", ())
```
mrava87: I may be on a tool old version of numpy locally (2.2.0)  but when I do

```
import numpy as np
a = np.ones(10)
view=a[5]
view.shape
```

it works and gives `()`. What is the motivation of using getattr here?

pylops_mpi/DistributedArray.py
```diff
@@ -644,7 +652,7 @@ def multiply(self, dist_array):
             ProductArray[:] = self.local_array * dist_array
         return ProductArray
 
-    def dot(self, dist_array):
+    def dot(self, dist_array, vdot=False):
```
mrava87: Can you please improve the docstring. Already before it would have been better to have it especially since this is not a private method but now it is also more important (even for us devs) not to forget what vdot does.

Also can you please add type hints (I know we are still a bit inconsistent and miss them in places here and there 😉

pylops_mpi/DistributedArray.py
```diff
@@ -656,8 +664,12 @@ def dot(self, dist_array):
         y = DistributedArray.to_dist(x=dist_array.local_array, base_comm=self.base_comm, base_comm_nccl=self.base_comm_nccl) \
             if self.partition in [Partition.BROADCAST, Partition.UNSAFE_BROADCAST] else dist_array
         # Flatten the local arrays and calculate dot product
+        if vdot:
```
mrava87: probably one liner is better?

`dot_func = ncp.vdot if vdot else ncp.dot`

pylops_mpi/DistributedArray.py
```diff
@@ -839,6 +851,15 @@ def ravel(self, order: Optional[str] = "C"):
         arr[:] = x
         return arr
 
+    def empty_like(self):
```
mrava87, line 867: mmh is this really the same as Numpy's one. In numpy:

```
x = np.ones(10)
y =  x.empty_like() # does NOT work
y = np.empty_like(x) # works
```

I am fine with doing
```
x = DistributedArray(...)
y = x..empty_like()
```

but since you said `similar to what we have in numpy.` I just wanted to raise this in case you did not think about. Feel free to ignore if you think the current approach is best or we could always turn it into a method like `local_split` if we think is more suitable..

pylops_mpi/DistributedArray.py
```diff
@@ -1038,13 +1059,13 @@ def multiply(self, stacked_array):
                 ProductArray[iarr][:] = (self[iarr] * stacked_array)[:]
         return ProductArray
 
-    def dot(self, stacked_array):
+    def dot(self, stacked_array, vdot=False):
         """Dot Product of Stacked Distributed Arrays
```
mrava87: Same comments of the other dot apply here

requirements-dev.txt
```diff
@@ -1,6 +1,6 @@
 numpy>=1.15.0
 scipy>=1.8.0
-pylops
+pylops>=2.6.0
```
mrava87: I am a bit reluctant about this as it is really a very recent version of PyLops. 

I think we probably should still fix this file as `pylops>=2.0` is what we really support (I am sure with pylops v1.x we would get some failures, but we don't care...), but going all the way to v2.6.0 to benefit from a minor feature in some of the solvers is maybe a bit too much?

environment-dev.yml
```diff
@@ -8,7 +8,7 @@ dependencies:
   - numpy>=1.15.0
   - scipy>=1.8.0
   - mpi4py
-  - pylops
+  - pylops>=2.6.0
```
mrava87: Again we should still add >=2.0.0 for consistency.

tutorials/reflectivity.py
```diff
@@ -0,0 +1,160 @@
...
+d_local = d_dist.local_array.reshape((ny_i, nx, nz))
+d = d_dist.asarray().reshape((ny, nx, nz))
+
+###############################################################################
+# We now perform sparsity-promotion inversion
+
+r0_dist = pylops_mpi.DistributedArray(global_shape=ny * nx * nz)
+r0_dist[:] = 0.
+
+rinv3d_dist = pylops_mpi.optimization.sparsity.ista(
```
mrava87, line 95: I'll add fista and a comparison of convergence.. but don't worry, I can make a separate PR so I don't hold you...

pylops_mpi/optimization/cls_sparsity.py
```diff
@@ -0,0 +1,724 @@
...
+        operates on a local NumPy/CuPy array and returns a thresholded array
+    thresh : :obj:`float`
+        Threshold value passed to ``threshf``
+
+    Returns
+    -------
+    x : :obj:`pylops_mpi.DistributedArray` or :obj:`pylops_mpi.StackedDistributedArray`
+        The input array after applying the thresholding operation.
+    """
+    if isinstance(x, DistributedArray):
```
mrava87, line 41: Nice 🚀

mrava87 (changes requested):
Thanks @rohanbabbar04. Changes look great!

Regarding your question about the setter method, I suspect you are right and may be due to the recent numpy and stricter handling of 0-dimensional arrays. So what you did looks good to me and I am quite sure it is backward compatible with older numpys that will also work with the old version of the code 😄 I just left one comment wrt to use of getattr.

Regarding `callback_stop`, I would suggest we take it out for now. It is already not added to the other solvers (which if we decided to keep, we should for consistency) and it is probably a too little of an incentive to force on every user such a recent version of PyLops for now.

I left a few minor comments, after those are handled I think we are good to go 🚀

requirements-dev.txt
rohanbabbar04: Agreed, have reverted it back to 2.0.0

tutorials/reflectivity.py
rohanbabbar04: Sure.

pylops_mpi/DistributedArray.py
rohanbabbar04: Yeah, I meant that only like it would be with `np.empty_like`, I should have framed it better 😅

pylops_mpi/DistributedArray.py
rohanbabbar04: Yup, you are right have changed it to view.shape

mrava87 (approved):

mrava87:
@rohanbabbar04 fantastic job 🚀

PR #195, opened 2026-05-08
Implement MPIFFT2D and MPIFFTND

rohanbabbar04:
- [x] Implement FFT using `mpi4py-fft`
- [x] Introduce `MPIFFTND`, `MPIFFT2D` and `_MPIBaseFFTND`
- [x] Basic Test to compare with pylops operators.
- [x] Working example in `examples\`
- [x] Update GA to include fftw libraries.
- [x] Add mpi implementations of `fftshift` and `ifftshift`.

**Note: `mpi4py-fft` works only with numpy arrays, and PFFT works with multi-dimensional arrays.**

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -36,10 +36,23 @@ jobs:
         uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python-version }}
+      - name: Install fftw libraries
+        run: |
+          case $(uname) in
+            Linux)
+              sudo apt update
+              sudo apt install -y -q libfftw3-dev
+              ;;
+            Darwin)
+              export HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK=1
+              brew install fftw
+              ;;
+          esac
       - name: Installing Dependencies
         run: |
           python -m pip install --upgrade pip setuptools
           if [ -f requirements.txt ]; then pip install -r requirements-dev.txt; fi
+          if [ -f requirements-fft.txt ]; then pip install -r requirements-fft.txt; fi
       - name: Install pylops-mpi
         run: pip install .
       - name: Testing using pytest-mpi
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -52,9 +52,8 @@ jobs:
         run: |
           python -m pip install --upgrade pip setuptools
           if [ -f requirements.txt ]; then pip install -r requirements-dev.txt; fi
-          if [ -f requirements-fft.txt ]; then pip install -r requirements-fft.txt; fi
       - name: Install pylops-mpi
-        run: pip install .
+        run: pip install .[all]
       - name: Testing using pytest-mpi
         run: |
           if [ "${{ matrix.mpi }}" = "openmpi" ]; then
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -21,5 +21,6 @@ dependencies:
   - nbsphinx
   - pydata-sphinx-theme
   - flake8
+  - mpi4py-fft
   - pip:
       - sphinx-gallery
--- a/examples/plot_ffts.py
+++ b/examples/plot_ffts.py
@@ -1,0 +1,111 @@
+"""
+Fourier Transform
+=================
+This example shows how to use the :py:class:`pylops_mpi.signalprocessing.MPIFFT2D`
+and :py:class:`pylops_mpi.signalprocessing.MPIFFTND` operators to apply the Fourier
+Transform to the model and the inverse Fourier Transform to the data.
+"""
+
+import matplotlib.pyplot as plt
+import numpy as np
+
+import pylops_mpi
+
+plt.close("all")
+
+###############################################################################
+# We start by applying the two dimensional MPI-distributed FFT to a
+# two-dimensional signal using :py:class:`pylops_mpi.signalprocessing.MPIFFT2D`.
+# The input signal is a :py:class:`pylops_mpi.DistributedArray` which is
+# distributed across MPI ranks before applying the transform.
+
+dt, dx = 0.005, 5
+nt, nx = 2**7, 2**8
+t = np.arange(nt) * dt
+x = np.arange(nx) * dx
+f0 = 10
+
+d = np.outer(np.sin(2 * np.pi * f0 * t), np.arange(nx) + 1)
+dist = pylops_mpi.DistributedArray.to_dist(x=d.ravel())
+
+FFTop = pylops_mpi.signalprocessing.MPIFFT2D(
+    dims=(nt, nx), sampling=(dt, dx)
+)
+
+D = FFTop * dist
+
+dinv = FFTop.H * D
+dinv = FFTop / D
+dinv = np.real(dinv.asarray()).reshape(nt, nx)
+
+D_2d = D.asarray().reshape(nt, nx)
+
+fig, axs = plt.subplots(2, 2, figsize=(10, 6))
+
+axs[0][0].imshow(d, vmin=-100, vmax=100, cmap="bwr")
+axs[0][0].set_title("Signal")
+axs[0][0].axis("tight")
+
+axs[0][1].imshow(
+    np.abs(np.fft.fftshift(D_2d, axes=1)[:nt // 2, :]), cmap="bwr"
+)
+axs[0][1].set_title("Fourier Transform")
+axs[0][1].axis("tight")
+
+axs[1][0].imshow(dinv, vmin=-100, vmax=100, cmap="bwr")
+axs[1][0].set_title("Inverted")
+axs[1][0].axis("tight")
+
+axs[1][1].imshow(d - dinv, vmin=-100, vmax=100, cmap="bwr")
+axs[1][1].set_title("Error")
+axs[1][1].axis("tight")
+
+fig.tight_layout()
+
+###############################################################################
+# We can also apply the three dimensional MPI-distributed FFT to a
+# three-dimensional signal using :py:class:`pylops_mpi.signalprocessing.MPIFFTND`.
+
+dt, dx, dy = 0.005, 5, 3
+nt, nx, ny = 2**7, 2**6, 13
+t = np.arange(nt) * dt
+x = np.arange(nx) * dx
+y = np.arange(ny) * dy
+f0 = 10
+
+d = np.outer(np.sin(2 * np.pi * f0 * t), np.arange(nx) + 1)
+d = np.tile(d[:, :, np.newaxis], [1, 1, ny])
+dist = pylops_mpi.DistributedArray.to_dist(x=d.ravel())
+
+FFTop = pylops_mpi.signalprocessing.MPIFFTND(
+    dims=(nt, nx, ny),
+    sampling=(dt, dx, dy)
+)
+
+D = FFTop * dist
+dinv = FFTop.H * D
+dinv = FFTop / D
+dinv = np.real(dinv.asarray()).reshape(nt, nx, ny)
+D_3d = D.asarray().reshape(nt, nx, ny)  # shape matches dims now
+
+fig, axs = plt.subplots(2, 2, figsize=(10, 6))
+
+axs[0][0].imshow(d[:, :, ny // 2], vmin=-20, vmax=20, cmap="bwr")
+axs[0][0].set_title("Signal")
+axs[0][0].axis("tight")
+axs[0][1].imshow(
+    np.abs(np.fft.fftshift(D_3d, axes=1)[:nx // 2, :, ny // 2]),
+    cmap="bwr"
+)
+axs[0][1].set_title("Fourier Transform")
+axs[0][1].axis("tight")
+
+axs[1][0].imshow(dinv[:, :, ny // 2], vmin=-20, vmax=20, cmap="bwr")
+axs[1][0].set_title("Inverted")
+axs[1][0].axis("tight")
+
+axs[1][1].imshow(d[:, :, ny // 2] - dinv[:, :, ny // 2], vmin=-20, vmax=20, cmap="bwr")
+axs[1][1].set_title("Error")
+axs[1][1].axis("tight")
+
+fig.tight_layout()
--- a/examples/plot_ffts.py
+++ b/examples/plot_ffts.py
@@ -34,7 +34,7 @@ FFTop = pylops_mpi.signalprocessing.MPIFFT2D(
 
 D = FFTop * dist
 
-dinv = FFTop.H * D
+dinv = FFTop / D
 dinv = np.real(dinv.asarray()).reshape(nt, nx)
 
 D_2d = D.asarray().reshape(nt, nx)
@@ -82,7 +82,7 @@ FFTop = pylops_mpi.signalprocessing.MPIFFTND(
 )
 
 D = FFTop * dist
-dinv = FFTop.H * D
+dinv = FFTop / D
 dinv = np.real(dinv.asarray()).reshape(nt, nx, ny)
 D_3d = D.asarray().reshape(nt, nx, ny)  # shape matches dims now
 
@@ -107,3 +107,4 @@ axs[1][1].set_title("Error")
 axs[1][1].axis("tight")
 
 fig.tight_layout()
+plt.show()
--- a/examples/plot_ffts.py
+++ b/examples/plot_ffts.py
@@ -107,4 +107,3 @@ axs[1][1].set_title("Error")
 axs[1][1].axis("tight")
 
 fig.tight_layout()
-plt.show()
--- a/examples/plot_ffts.py
+++ b/examples/plot_ffts.py
@@ -34,7 +34,7 @@ FFTop = pylops_mpi.signalprocessing.MPIFFT2D(
 
 D = FFTop * dist
 
-dinv = FFTop / D
+dinv = FFTop.H * D
 dinv = np.real(dinv.asarray()).reshape(nt, nx)
 
 D_2d = D.asarray().reshape(nt, nx)
@@ -82,7 +82,7 @@ FFTop = pylops_mpi.signalprocessing.MPIFFTND(
 )
 
 D = FFTop * dist
-dinv = FFTop / D
+dinv = FFTop.H * D
 dinv = np.real(dinv.asarray()).reshape(nt, nx, ny)
 D_3d = D.asarray().reshape(nt, nx, ny)  # shape matches dims now
 
--- a/examples/plot_ffts.py
+++ b/examples/plot_ffts.py
@@ -34,7 +34,6 @@ FFTop = pylops_mpi.signalprocessing.MPIFFT2D(
 
 D = FFTop * dist
 
-dinv = FFTop.H * D
 dinv = FFTop / D
 dinv = np.real(dinv.asarray()).reshape(nt, nx)
 
@@ -83,7 +82,6 @@ FFTop = pylops_mpi.signalprocessing.MPIFFTND(
 )
 
 D = FFTop * dist
-dinv = FFTop.H * D
 dinv = FFTop / D
 dinv = np.real(dinv.asarray()).reshape(nt, nx, ny)
 D_3d = D.asarray().reshape(nt, nx, ny)  # shape matches dims now
--- a/pylops_mpi/signalprocessing/FFT2D.py
+++ b/pylops_mpi/signalprocessing/FFT2D.py
@@ -41,6 +41,21 @@ class MPIFFT2D(MPIFFTND):
         model is real. Note that the real FFT is applied only to the first
         dimension to which the FFT2D operator is applied (last element of
         ``axes``)
+    ifftshift_before : :obj:`tuple` or :obj:`bool`, optional
+        Apply ifftshift (``True``) or not (``False``) to model vector (before FFT).
+        Consider using this option when the model vector's respective axis is symmetric
+        with respect to the zero value sample. This will shift the zero value sample to
+        coincide with the zero index sample. With such an arrangement, FFT will not
+        introduce a sample-dependent phase-shift when compared to the continuous Fourier
+        Transform. When passing a single value, the shift will the same for every direction.
+        Pass a tuple to specify which dimensions are shifted.
+    fftshift_after : :obj:`tuple` or :obj:`bool`, optional
+        Apply fftshift (``True``) or not (``False``) to data vector (after FFT).
+        Consider using this option when you require frequencies to be arranged
+        naturally, from negative to positive. When not applying fftshift after FFT,
+        frequencies are arranged from zero to largest positive, and then from negative
+        Nyquist to the frequency bin before zero. When passing a single value, the shift
+        will the same for every direction. Pass a tuple to specify which dimensions are shifted.
     dtype : :obj:`str`, optional
         Type of elements in input array. Note that the ``dtype`` of the operator
         is the corresponding complex type even when a real type is provided.
@@ -116,6 +131,8 @@ class MPIFFT2D(MPIFFTND):
         sampling: float | Sequence[float] = 1.0,
         norm: str = "none",
         real: bool = False,
+        ifftshift_before: bool = False,
+        fftshift_after: bool = False,
         dtype: DTypeLike = "complex128",
         base_comm: MPI.Comm = MPI.COMM_WORLD,
         **kwargs_fft,
@@ -127,8 +144,9 @@ class MPIFFT2D(MPIFFTND):
         if len(axes) != 2:
             msg = "FFT2D must be applied along exactly two dimensions"
             raise ValueError(msg)
-        super().__init__(dims=dims, axes=axes, sampling=sampling, norm=norm, real=real,
-                         dtype=dtype, base_comm=base_comm, **kwargs_fft)
+        super().__init__(dims=dims, axes=axes, sampling=sampling, norm=norm, real=real, dtype=dtype,
+                         ifftshift_before=ifftshift_before, fftshift_after=fftshift_after, base_comm=base_comm,
+                         **kwargs_fft)
         self.f1, self.f2 = self.fs
         del self.fs
 
--- a/pylops_mpi/signalprocessing/FFT2D.py
+++ b/pylops_mpi/signalprocessing/FFT2D.py
@@ -1,0 +1,139 @@
+from typing import Sequence
+
+from mpi4py import MPI
+
+from pylops.utils import DTypeLike, InputDimsLike
+
+from pylops_mpi.DistributedArray import DistributedArray
+from pylops_mpi.signalprocessing.FFTND import MPIFFTND
+
+
+class MPIFFT2D(MPIFFTND):
+    r"""Two-dimensional Fast-Fourier Transform.
+
+    Apply two-dimensional Fast-Fourier Transform (FFT) to any pair of ``axes`` of a
+    multidimensional array.
+
+    When using ``real=True``, the result of the forward is also multiplied by
+    :math:`\sqrt{2}` for all frequency bins except zero and Nyquist, and the input of
+    the adjoint is multiplied by :math:`1 / \sqrt{2}` for the same frequencies.
+
+    For a real valued input signal, it is advised to use the flag ``real=True``
+    as it stores the values of the Fourier transform of the last axis in ``axes`` at positive
+    frequencies only as values at negative frequencies are simply their complex conjugates.
+
+    Parameters
+    ----------
+    dims : :obj:`tuple`
+        Number of samples for each dimension
+    axes : :obj:`tuple`, optional
+        Pair of axes along which FFT2D is applied
+    sampling : :obj:`tuple` or :obj:`float`, optional
+        Sampling steps for each axis in ``axes``. When supplied a single value, it is used
+        for both axes.
+    norm : `{"none", "1/n"}`, optional
+        - "none": Does not scale the forward or the adjoint FFT transforms. Default is "none".
+        - "1/n": Scales both the forward and adjoint FFT transforms by
+          :math:`1/N_F`.
+    real : :obj:`bool`, optional
+        Model to which fft is applied has real numbers (``True``) or not
+        (``False``). Used to enforce that the output of adjoint of a real
+        model is real. Note that the real FFT is applied only to the first
+        dimension to which the FFT2D operator is applied (last element of
+        ``axes``)
+    dtype : :obj:`str`, optional
+        Type of elements in input array. Note that the ``dtype`` of the operator
+        is the corresponding complex type even when a real type is provided.
+        In addition, note that the NumPy backend does not support returning ``dtype``
+        different from ``complex128``.
+    base_comm : :obj:`mpi4py.MPI.Comm`, optional
+        MPI Base Communicator. Defaults to ``mpi4py.MPI.COMM_WORLD``.
+    **kwargs_fft
+        Arbitrary keyword arguments to be passed to the selected fft method
+
+    Attributes
+    ----------
+    f1 : :obj:`numpy.ndarray`
+        Discrete Fourier Transform sample frequencies along ``axes[0]``
+    f2 : :obj:`numpy.ndarray`
+        Discrete Fourier Transform sample frequencies along ``axes[1]``
+    nffts : :obj:`tuple` or :obj:`int`, optional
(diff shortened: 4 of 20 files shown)


rohanbabbar04:
Worked on implementing `fft` using `mpi4py-fft`. Overall, the implementation is straightforward — getting the data into PFFT and retrieving it back into our `Distributed Array`. There are some key considerations to keep in mind.

- Since we have `axes` as a parameter in the class, we need to ensure that `x` (the data) is redistributed across the `axes[0]` axis before pushing it into the `DistArray` (`mpi4py-fft`). This is a key step which acts as the starting point.
- I tried implementing `fftshift` and `ifftshift`. I think the best approach is: if the distribution axis differs from the required axis, we can compute it directly; otherwise, we redistribute to a new axis first and then compute.
- PFFT doesn't support 1-D arrays, which makes sense since there is no free axis available for computation. We can explore a workaround for this.
- The `nffts` parameter from the PyLops version is currently missing. PFFT does have a padding parameter, but we will need to investigate how it can be mapped to `nffts`.
- `mpi4py-fft` works with NumPy arrays, since `DistArray` is built on top of `np.ndarray`. For CuPy/NCCL support, we should look into using `nvidia.distributed.fft`.

Once we agree on the implementation, I will go ahead and update the documentation including the `mpi4py-fft`.

examples/plot_ffts.py
```diff
@@ -0,0 +1,111 @@
...
+dist = pylops_mpi.DistributedArray.to_dist(x=d.ravel())
+
+FFTop = pylops_mpi.signalprocessing.MPIFFT2D(
+    dims=(nt, nx), sampling=(dt, dx)
+)
+
+D = FFTop * dist
+
+dinv = FFTop.H * D
+dinv = FFTop / D
```
mrava87, line 37: I would remove this as .H is already inv for fit

examples/plot_ffts.py
```diff
@@ -0,0 +1,111 @@
...
+# distributed across MPI ranks before applying the transform.
+
+dt, dx = 0.005, 5
+nt, nx = 2**7, 2**8
+t = np.arange(nt) * dt
+x = np.arange(nx) * dx
+f0 = 10
+
+d = np.outer(np.sin(2 * np.pi * f0 * t), np.arange(nx) + 1)
+dist = pylops_mpi.DistributedArray.to_dist(x=d.ravel())
```
mrava87, line 29: Starting my review from this example. I assume that by raveling d, we are here showing what we would do when we solve an inverse problem, so d is technically 2d but its flattened and so distributed over the first/slow/t axis?

Question: for the sake of this small example, could we show that the operator `MPIFFT2D` also works for 2D Distributed arrays - eg `dist = pylops_mpi.DistributedArray.to_dist(x=d, axis=0)` and `dist = pylops_mpi.DistributedArray.to_dist(x=d, axis=1) `

One of the reasons I am asking is that I have been pondering for a while if we could support actual NDarrays also in the solvers like we do with PyLops - this is a nice feature for users as they can just think ND, not flattened-ND... internally in PyLops we just flatten at the start and reshape at the end (https://github.com/PyLops/pylops/blob/4ca81109f504072273a2a94c858bbc3860854abf/pylops/utils/decorators.py#L38), maybe something like this done on the local_array could work here too..

mrava87: But even if this is not the case for now, it would still be nice to showcase an example where we operate on actual distributed Ndarrays 😄

pylops_mpi/signalprocessing/FFTND.py
```diff
@@ -0,0 +1,283 @@
...
+    MPIFFT2D: Two-dimensional FFT
+
+    Raises
+    ------
+    ValueError
+        - If ``norm`` is not one of "none", or "1/n".
+
+    Notes
+    -----
+    The MPIFFTND operator (using ``norm="none"``) applies the N-dimensional forward
```
mrava87: I did a quick check of the documentation and in general our approach to operators we re-implement in pylops-mpi has been to focus the notes on the aspects related to the distributed version of the algorithm used rather than the general operator... this is really a copy-paste from the PyLops operator that is not very useful because if one wants to understand what is `FFT2D` they can just look at PyLops's documentation. What really matters is to explain how its **MPI**`FFT2D` version differ for it.

This leads me to the main question of this PR: could you explain the overall rationale you took in the current implementation. So far this

```
# Distribute only along axes[0]; all other axes are non-distributed (0=distributed, 1=not)
subcomm_dims = (np.arange(len(axes)) != axes[0]).astype(int)
```

makes me think we can only distribute axis=0 and so to some extent (I guess) only pass 1D DistributedArray of ND but with axis=0?

environment-dev.yml
mrava87, line 1: I think we could perhaps add a section with optional dependencies in pyproject.toml like in https://github.com/PyLops/pylops/blob/4ca81109f504072273a2a94c858bbc3860854abf/pyproject.toml#L39 and add `mpi4py-fft` to that list.

PS: the approach we took in PyLops until recently was to keep out any GPU-based library (as it was difficult to set at pyproject.toml level but to add all others... recently, once we switched to `uv` we have been able to add also cupy/torch, we can at some point try to do the same here...

pylops_mpi/signalprocessing/FFTND.py
```diff
@@ -0,0 +1,283 @@
...
+
+    @reshaped
+    def _matvec(self, x: DistributedArray) -> DistributedArray:
+        if x.engine == "cupy":
+            raise ValueError(f"x should be a numpy array with engine=numpy"
+                             f"Got  {x.engine} instead...")
+        if x.partition != Partition.SCATTER:
+            raise ValueError(f"x should have partition={Partition.SCATTER}"
+                             f"Got  {x.partition} instead...")
+        ncp = get_array_module(x.local_array)
```
mrava87: mmh why is this needed if we already prevent to get cupy arrays through (first check)?

pylops_mpi/signalprocessing/FFTND.py
```diff
@@ -0,0 +1,283 @@
...
+        if self.ifftshift_before.any():
+            x = ifftshift_nd(x, axes=self.axes[self.ifftshift_before])
+        if not self.clinear:
+            x[:] = ncp.real(x.local_array)
+        # Allocate distributed arrays for input and output
+        u_dist = newDistArray(self.fft, forward_output=False)
```
mrava87: I would prefer to call these x/y... so maybe `x_dist`/`y_dist`

pylops_mpi/signalprocessing/FFTND.py
```diff
@@ -0,0 +1,283 @@
...
+
+        self._kwargs_fft = kwargs_fft
+        if self.norm is _FFTNorms.NONE:
+            self._scale = np.prod(self.nffts)
+        elif self.norm is _FFTNorms.ONE_OVER_N:
+            self._scale = 1.0 / np.prod(self.nffts)
+        fft_dtype = self.rdtype if self.real else self.cdtype
+        # Distribute only along axes[0]; all other axes are non-distributed (0=distributed, 1=not)
+        subcomm_dims = (np.arange(len(axes)) != axes[0]).astype(int)
+        subcomm = Subcomm(self.base_comm, dims=np.resize(subcomm_dims, len(dims)))
```
mrava87: A bit lost about these two lines... what are trying to achieve?

If I understand your comment, I would think this is a much easier way to achieve it:
```
subcomm_dims = np.ones(len(dims))
subcomm_dims[axes[0]] = 1
```

Apart from simplifying the code, can you explain why you want to only distribute along axes[0] instead of for example along the axis that the input is actually distributed... just trying to think if we can minimize any re-distribution. Of course, this way you would need to ask upfront which axis the input that the operator expects will be distributed, but if that is the price to pay I would still prefer it to doing more re-distributions than needed?

pylops_mpi/signalprocessing/FFTND.py
```diff
@@ -0,0 +1,283 @@
...
+                             f"Got  {x.partition} instead...")
+        ncp = get_array_module(x.local_array)
+        if self.ifftshift_before.any():
+            x = ifftshift_nd(x, axes=self.axes[self.ifftshift_before])
+        if not self.clinear:
+            x[:] = ncp.real(x.local_array)
+        # Allocate distributed arrays for input and output
+        u_dist = newDistArray(self.fft, forward_output=False)
+        u_hat = newDistArray(self.fft, forward_output=True)
+        # Redistribute input to match the axis decomposed by PFFT
```
mrava87: Linked to what I asked in my previous comment... what is the main rational behind this?

pylops_mpi/signalprocessing/FFTND.py
```diff
@@ -0,0 +1,283 @@
...
+        self._kwargs_fft = kwargs_fft
+        if self.norm is _FFTNorms.NONE:
+            self._scale = np.prod(self.nffts)
+        elif self.norm is _FFTNorms.ONE_OVER_N:
+            self._scale = 1.0 / np.prod(self.nffts)
+        fft_dtype = self.rdtype if self.real else self.cdtype
+        # Distribute only along axes[0]; all other axes are non-distributed (0=distributed, 1=not)
+        subcomm_dims = (np.arange(len(axes)) != axes[0]).astype(int)
+        subcomm = Subcomm(self.base_comm, dims=np.resize(subcomm_dims, len(dims)))
+        self.fft = PFFT(subcomm, self.dims, axes=self.axes, dtype=fft_dtype, collapse=True, **self._kwargs_fft)
```
mrava87: How do you handle `real=True`? 

In PyLops we call, from what I see here https://mpi4py-fft.readthedocs.io/en/latest/parallel.html, it seems like you should pass something like `transforms = {axes: (rfftn, irfftn)}` to `PFFT` but you don't seem to?

requirements-fft.txt
mrava87, line 1: Why do you need a separate file?

.github/workflows/build.yml
```diff
@@ -36,10 +36,23 @@ jobs:
         uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python-version }}
+      - name: Install fftw libraries
```
mrava87, line 39: Do you need this? In pylops we simply install `pyfftw`, would that not be enough also for `mpi4py-fft` to work?

mrava87 (changes requested):
@rohanbabbar04 this is a great addition to pylops-mpi!

I have gone through the PR mostly trying to understand the rationale of your code and the decision made in the implementation - left some comments and questions?

Once we agree on the right approach, I'll do a more focused review on the actual code 😄

pylops_mpi/signalprocessing/FFTND.py
```diff
@@ -0,0 +1,283 @@
...
+        u_hat = newDistArray(self.fft, forward_output=True)
+        # Redistribute input to match the axis decomposed by PFFT
+        x = x.redistribute(axis=self.axes[0])
+        u_dist[:] = x.local_array
+        # Perform the parallel forward FFT
+        self.fft.forward(u_dist, u_hat, normalize=False)
+
+        # Axis along which PFFT decomposes the output array across MPI processes
+        dist_axis = [i for i, s in enumerate(u_hat.subcomm) if s.Get_size() > 1]
+        y = DistributedArray(global_shape=self.dimsd, dtype=self.dtype, axis=dist_axis[0] if dist_axis else 0,
```
mrava87: Does this mean that the output could have a different distribution of the input?

pylops_mpi/signalprocessing/FFTND.py
```diff
@@ -0,0 +1,283 @@
...
+                             base_comm=x.base_comm, base_comm_nccl=x.base_comm_nccl, engine=x.engine)
+        y[:] = u_hat
+        if self.real:
+            # Redistribute so that self.axes[-1] is not the one sliced
+            safe_axis = next(i for i in range(len(self.dims)) if i != self.axes[-1])
+            y = y.redistribute(axis=safe_axis)
```
mrava87: But then here, the axis over which the array is distributed can change?

PS: I think in PyLops we used this code pattern as it is cheap (view-only)
```
y = ncp.swapaxes(y, -1, self.axes[-1])
y[..., 1 : 1 + (self.nffts[-1] - 1) // 2] *= ncp.sqrt(2)
y = ncp.swapaxes(y, self.axes[-1], -1)
```

but here to avoid redistribution we can probably use slices to define which axis should get the `*= ncp.sqrt(2)` instead of redistribution + swap to last axis + swap back?

pylops_mpi/signalprocessing/FFTND.py
rohanbabbar04: Ah!
You can actually control this using the `dtype` parameter, for `fft` it should be a complex-datatype.

pylops_mpi/signalprocessing/FFTND.py
rohanbabbar04: Yeah, I was working on something, and I think I was able to make it work using just slices.

.github/workflows/build.yml
rohanbabbar04: Yeah, it should be installed. It should be required for `pyfftw` as well, which is what their [documentation](https://github.com/pyfftw/pyfftw#requirements-ie-what-it-was-designed-for) refers to. I never tried installing `pyfftw` without fftw, but we surely can try that. 
Similarly, `mpi4py-fft` requires these [libraries](https://github.com/mpi4py/mpi4py-fft#installation) to install.

requirements-fft.txt
rohanbabbar04: Yup, removed it...moved it to optional-dependencies.

environment-dev.yml
rohanbabbar04: Yup, have added it as an optional dependency. Will add some documentation around it, once we approve this :)

pylops_mpi/signalprocessing/FFTND.py
rohanbabbar04: Yup, removed it

pylops_mpi/signalprocessing/FFTND.py
rohanbabbar04: Done.

.github/workflows/build.yml
mrava87: Mmh could you try removing this as see what happens... in PyLops we do not add any specific installation of FFTW to the CI (https://github.com/PyLops/pylops/blob/v2.6.0/.github/workflows/build.yaml) and simply pip install pyfftw and it works. From what ChatGPT tells me, the pip wheel may bundle the FFTW3 binaries (https://chatgpt.com/share/6a077b8f-fb60-83eb-8c9a-105c2398ab75)

I rather not complicate the CI unless it's really needed.

rohanbabbar04: Yeah, I actually did it, and encountered this error https://github.com/PyLops/pylops-mpi/actions/runs/25574058834/job/75076724575. Post which I looked into the `mpi4py-fft` actions, and added the run here.

pylops_mpi/signalprocessing/FFTND.py
rohanbabbar04: @mrava87, this is what I thought just to make the function pretty flexible.

This `subcomm` parameter controls which axis should be distributed. By default, `PFFT` determines this automatically based on the number of ranks (and the distribution type: slab or non-slab). Also, our Distributed Array supports Slab(distribution along one axis).

I changed this to `subcomm_dim[0] = 0`, meaning only the first axis is distributed — which is what our reshaped method does. However, if the last axis provided by the user (via the axes param) is 0, we need to override this, as `PFFT` raises an error in that case(changing this to `subcomm_dim[1]` = 0).

The distribution axis for the forward and backward functions depends on the axes argument provided by the user. To handle this, I introduced two internal parameters — `_pfft_in_axis` and `_pfft_out_axis` — which track the distribution axis of the input and output respectively. These differ because `PFFT` may internally transpose the axis during FFT computation, meaning the output of forward (and similarly backward) can be distributed across different axes.

examples/plot_ffts.py
rohanbabbar04: Yeah, this is what I had in my mind as well. We would need to tweak some things in the main Linear Operator class (to allow nd dist arrays). After that, maybe some operators which can easily support nd arrays, we can start supporting that. I think FFT and Derivatives would be a good start.

mrava87: Ok, it makes sense 😄 as of today, we can't even apply the operator to an actual N-dimensional DistributedArray so what you do is the only thing we can show

But, yes, it would be nice to explore if we could allow any MPILinearOperator to also work with N-dimensional DistributedArrays like PyLops LinearOperator can work with numpy arrays whose shape matches the operator dims/dimsd.

pylops_mpi/signalprocessing/FFTND.py
```diff
@@ -0,0 +1,302 @@
...
+
+    When ``norm="none"``, the adjoint multiplies by :math:`N_F` to cancel this internal scaling,
+    returning a true unscaled adjoint. The result is then flattened back to a 1D
+    :class:`pylops_mpi.DistributedArray`. All inter-rank data movement is handled internally by
+    ``mpi4py_fft``.
+    """
+    def __init__(
+        self,
+        dims: InputDimsLike,
+        axes: InputDimsLike = (0, 1, 2),
```
mrava87, line 157: I just realized you don't pass `nffts`, so this is eventually always equal None and so set to be equal to dims, any reason why?

pylops_mpi/signalprocessing/FFT2D.py
```diff
@@ -0,0 +1,163 @@
...
+        decomposition, transform axes, and dtype.
+
+    See Also
+    --------
+    MPIFFTND: N-dimensional FFT
+
+    Raises
+    ------
+    ValueError
+        - If ``norm`` is not one of "none", or "1/n".
```
mrava87, line 102: Remove `-` if ValueError is raised only for one case

mrava87: But looking at the init, there are at least 2 raises... fix this.

pylops_mpi/signalprocessing/FFT2D.py
```diff
@@ -0,0 +1,163 @@
...
+
+    Notes
+    -----
+    The MPIFFT2D operator performs forward and adjoint passes on a
```
mrava87: `The MPIFFT2D operator performs forward and adjoint passes on a`  -> `The MPIFFT2D operator applies the forward and inverse 2-dimensional Fast Fourier transform to a`

pylops_mpi/signalprocessing/_baseffts.py
```diff
@@ -0,0 +1,169 @@
...
+            msg = (
+                "`axes`, `nffts`, `sampling`, `ifftshift_before` and "
+                "`fftshift_after` must the have same number of elements. Received "
+                f"{self.naxes}, {len(self.nffts)}, {len(self.sampling)}, "
+                f"{len(self.ifftshift_before)} and {len(self.fftshift_after)}, "
+                "respectively."
+            )
+            raise ValueError(msg)
+
+        # Check if the user provided nfft smaller than n. See _BaseFFT for
```
mrava87: This comment does not make sense right now, as a user is not even allowed to pass nffts (in the public API classes)...

mrava87: Also what is _BaseFFT here?

pylops_mpi/signalprocessing/FFT2D.py
```diff
@@ -0,0 +1,163 @@
...
+    :class:`pylops_mpi.DistributedArray`, which is internally reshaped to the 2-dimensional layout
+    defined by ``dims``. The 2-dimensional FFT is then applied across MPI ranks using ``mpi4py_fft``'s
+    :class:`mpi4py_fft.mpifft.PFFT` class, with the global array decomposed via a pencil decomposition.
+    :class:`mpi4py_fft.pencil.Subcomm` selects the axis of distribution: ``axis=0`` by default,
```
mrava87:
```
:class:`mpi4py_fft.pencil.Subcomm` selects the axis of distribution: ``axis=0`` by default,
shifting to ``axis=1`` if ``axes[-1] == 0`` to avoid a conflict between the transform and decomposition axes.
```

This is still unclear. Since MPI operators can only be applied to 1D DistributedArray, i guess we always assume that the first axis of the underlying 2d array is distributed, so when internally the local array is reshaped, the only axis over which a straighforward fft isnt allow is axis=0. 

Now if axes=(0,1) this is all fine, right? But if axes=(1,0), the Subcomm is defined such that axis=1 is assumed to be distributed but then we need to use `.redistribute` to make this happen? And then at the end after the FFT is applied we let @reshaped to bring it back to 1D?

Not asking to write in such detail but things like `to avoid a conflict between the transform and decomposition axes.` are very unclear and a user would not gain much from it

pylops_mpi/signalprocessing/FFTND.py
mrava87: Better, but I still left some comments

pylops_mpi/signalprocessing/FFTND.py
```diff
@@ -0,0 +1,302 @@
...
+    ------
+    ValueError
+        - If ``norm`` is not one of "none", or "1/n".
+
+    Notes
+    -----
+    The MPIFFTND operator performs forward and adjoint passes on a
+    :class:`pylops_mpi.DistributedArray`, which is internally reshaped to the N-dimensional layout
+    defined by ``dims``. The N-dimensional FFT is then applied across MPI ranks using ``mpi4py_fft``'s
+    :class:`mpi4py_fft.mpifft.PFFT` class, with the global array decomposed via a pencil decomposition.
```
mrava87: Is it not slab? I thought we dont distribute over 2 axes?

Basically left of the 2 figures in https://mpi4py-fft.readthedocs.io/en/latest/global.html# not right?

pylops_mpi/signalprocessing/FFTND.py
```diff
@@ -0,0 +1,302 @@
...
+                "To respect the passed dtype, data will be cast to {self.cdtype}.",
+                stacklevel=2,
+            )
+        if self.norm is _FFTNorms.NONE:
+            self._scale = np.prod(self.nffts)
+        elif self.norm is _FFTNorms.ONE_OVER_N:
+            self._scale = 1.0 / np.prod(self.nffts)
+        fft_dtype = self.rdtype if self.real else self.cdtype
+        subcomm_dims = np.ones(len(dims), dtype=int)
+        # axis=0 for the initial distribution by default, if the final axis is 0, distribute along axis 1 instead.
```
mrava87: `if the final axis is 0`, maybe say `if the first axis over which FFT is applied is axis=0, the input array is first redistributed over axis=1 prior to applying FFT`

pylops_mpi/signalprocessing/FFTND.py
```diff
@@ -0,0 +1,302 @@
...
+        if axes[-1] == 0:
+            subcomm_dims[1] = 0
+        else:
+            subcomm_dims[0] = 0
+        subcomm = Subcomm(base_comm, subcomm_dims)
+        self.fft = PFFT(subcomm, self.dims, axes=self.axes, dtype=fft_dtype)
+        self._pfft_in_axis = next(
```
mrava87, line 205: This is still very cryptic, please add MANY MORE comments to your code in hard-to-understand areas.

pylops_mpi/signalprocessing/FFTND.py
mrava87: Ok I think I get it... we apply reshaped so we always create internally a Nd distributed array with scattering over axes=0 but PFFT does not like the first axis over which the FFT is computed to be distributed, so if this is the case we redistribute over another axis?

pylops_mpi/signalprocessing/FFTND.py
mrava87: I guess I get it now...

.github/workflows/build.yml
mrava87: Ok I saw the same locally... seems like that `pyfftw` installs from PyPI in a self-consistent manner (no need for FFTW) but mpi4py-fft doesn't, hence the need for FFTW installation upfront

mrava87:
@rohanbabbar04 thanks for the updates!

I left a few additional comments (and unresolved some conversations to make sure I fully undestand some of your design choices... should be clear now but please confirm).

Once you have handled these last few points, I think this is good to go!

Next, we can look into enabling N-D Distributed arrays into MPI Linear Operators as in some cases like FFT we could benefit from not having to always distribute the first axis... I had a little play and it seems doable with minor code changes, I may create a draft PR to document what I did so far

pylops_mpi/signalprocessing/FFTND.py
rohanbabbar04: Yes, you are right...
In the future, If the person provides Nd arrays along `axis=0`, we wouldn't need to redistribute and straightaway we can perform the fft...but if the user let's say has `axis=-1`, as the distribution axis, a minor redistribution before we perform fft (this will make the function more flexible).

I would suggest we keep the `self.fft` fixed rather than changing it with every `matvec` or `rmatvec` call (so we initialise it once inside the init itself.)

examples/plot_ffts.py
rohanbabbar04: I agree, I can open a new issue for this, and I will start working on this 😀

mrava87: Sounds good :)

pylops_mpi/signalprocessing/FFTND.py
mrava87: Yes makes sense to fix the `self.fft` object 😄 which basically means (like for many PyLops operators) once a operator is defined one can't change the approach of for example how the input distributed array is distributed....

pylops_mpi/signalprocessing/FFTND.py
rohanbabbar04: Yes, it could be, as it depends on the axes parameter.

rohanbabbar04:
> @rohanbabbar04 thanks for the updates!
> 
> I left a few additional comments (and unresolved some conversations to make sure I fully undestand some of your design choices... should be clear now but please confirm).
> 
> Once you have handled these last few points, I think this is good to go!
> 
> Next, we can look into enabling N-D Distributed arrays into MPI Linear Operators as in some cases like FFT we could benefit from not having to always distribute the first axis... I had a little play and it seems doable with minor code changes, I may create a draft PR to document what I did so far

Thanks @mrava87 for the review, I have added the comments to the respective questions. Do let me know if you need any more information 🙂 ?

pylops_mpi/signalprocessing/FFTND.py
mrava87: @rohanbabbar04 ok, this is not ideal as it is not really predictable. 

A user having a global 2d array chunked over axis=0 and flattened and then put into a 1D distributed array may get back a new 1D distributed array that contains local arrays that if reshaped to 2d have chunks over axis=1 and entire axis 0?

If so, how do you see a user knowing this, is there at least a parameter in the FFT operator they can consult?

mrava87:
> > @rohanbabbar04 thanks for the updates!
> > I left a few additional comments (and unresolved some conversations to make sure I fully undestand some of your design choices... should be clear now but please confirm).
> > Once you have handled these last few points, I think this is good to go!
> > Next, we can look into enabling N-D Distributed arrays into MPI Linear Operators as in some cases like FFT we could benefit from not having to always distribute the first axis... I had a little play and it seems doable with minor code changes, I may create a draft PR to document what I did so far
> 
> Thanks @mrava87 for the review, I have added the comments to the respective questions. Do let me know if you need any more information 🙂 ?

Quite a few comments unanswered (you may have missed completely my comments from the second review I did... I pinged you in each of them, so you can easily find them 😄

pylops_mpi/signalprocessing/FFTND.py
mrava87: @rohanbabbar04

pylops_mpi/signalprocessing/FFTND.py
mrava87: @rohanbabbar04

pylops_mpi/signalprocessing/FFT2D.py
mrava87: @rohanbabbar04

pylops_mpi/signalprocessing/FFT2D.py
mrava87: @rohanbabbar04

pylops_mpi/signalprocessing/_baseffts.py
mrava87: @rohanbabbar04

pylops_mpi/signalprocessing/FFT2D.py
mrava87: @rohanbabbar04

pylops_mpi/signalprocessing/FFTND.py
mrava87: @rohanbabbar04

pylops_mpi/signalprocessing/FFTND.py
mrava87: @rohanbabbar04

pylops_mpi/signalprocessing/FFTND.py
mrava87: @rohanbabbar04

mrava87 (changes requested):

pylops_mpi/signalprocessing/FFT2D.py
rohanbabbar04: Done

pylops_mpi/signalprocessing/FFT2D.py
rohanbabbar04: Done

pylops_mpi/signalprocessing/FFTND.py
rohanbabbar04: Yes, I currently need to determine the best way to handle setting the FFT output shape (`self.nffts`) in `PFFT`. I plan to work on this in the next phase when introducing this parameter. To avoid confusion, I have also removed the parameter from `_MPIBaseFFTND` for now.

pylops_mpi/signalprocessing/_baseffts.py
rohanbabbar04: Yes, I have removed the parameter, and just setting it to be `dims`. :sweat_smile:

pylops_mpi/signalprocessing/FFT2D.py
rohanbabbar04: Yup, I have changed it up a bit, more focus on what will happen, especially focusing on the sub_comm, redistribute and PFFT.

pylops_mpi/signalprocessing/FFTND.py
rohanbabbar04: Yes, I meant the pencil object provided by PFFT, but yeah it still looks strange, so I have made the change to mention `slab` specifically (...that sounds much better).

pylops_mpi/signalprocessing/FFTND.py
rohanbabbar04: Yeah, I agree we should be adding some comments.I have added some comments explaining the the axis definition.

pylops_mpi/signalprocessing/FFTND.py
mrava87: Sounds good. Fine with hav ing a first version with fixed `nffts` as long as it is consistent across all classes like you did now

examples/plot_ffts.py
```diff
@@ -107,3 +107,4 @@
 axs[1][1].axis("tight")
 
 fig.tight_layout()
+plt.show()
```
mrava87: Not needed, I removed it.

mrava87 (approved):
Thanks @rohanbabbar04, this looks great now.

Happy with `nfft` being addressed later in a second PR, and perhaps we will also revisit things a bit when we will be able to pass ND-distributed arrays to solvers... right now I think what you do with internal redistributions and final reflattening with distribution over the axis=0 is the only possible way 😄 

I will merge once the CI is finished.

examples/plot_ffts.py
rohanbabbar04: Oops, thanks...
I was testing it and forgot to remove it 😅 .

PR #199, opened 2026-06-18
Add `dims` and `dimsd` to MPI Operators

rohanbabbar04:
- Introduce `dims` and `dimsd` as parameters in `MPILinearOperator` and `MPIStackedLinearOperator`.
- Add `dims` and `dimsd` to operators.

### DIFF
--- a/pylops_mpi/LinearOperator.py
+++ b/pylops_mpi/LinearOperator.py
@@ -1,6 +1,8 @@
+from __future__ import annotations
+
 import numpy as np
 from mpi4py import MPI
-from typing import Callable, Optional
+from typing import Callable, Optional, Union
 
 from scipy.sparse._sputils import isintlike
 from scipy.sparse.linalg._interface import _get_dtype
@@ -26,15 +28,19 @@ class MPILinearOperator:
     Note that whilst this operator could also be used with different
     :obj:`pylops.LinearOperator` across ranks, and with a
     :class:`pylops_mpi.DistributedArray` with ``Partition.SCATTER``, it is however
-    reccomended to use the :class:`pylops_mpi.basicoperators.MPIBlockDiag` operator
+    recommended to use the :class:`pylops_mpi.basicoperators.MPIBlockDiag` operator
     instead as this can also handle distributed arrays with subcommunicators.
 
     Parameters
     ----------
-    Op : :obj:`pylops.LinearOperator`, optional
-        PyLops Linear Operator to wrap. Defaults to ``None``.
+    Op : :obj:`pylops.LinearOperator` or :obj:`pylops_mpi.MPILinearOperator`, optional
+        If other arguments are provided, they will overwrite those obtained from ``Op``. Defaults to ``None``.
     shape : :obj:`tuple(int, int)`, optional
-        Shape of the MPI Linear Operator. Defaults to ``None``.
+        Shape of the MPI Linear Operator. If not provided, obtained from ``dims`` and ``dimsd``.
+    dims : :obj:`tuple(int, ..., int)`, optional
+        Dimensions of model. If not provided, ``(self.shape[1],)`` is used.
+    dimsd : :obj:`tuple(int, ..., int)`, optional
+        Dimensions of data. If not provided, ``(self.shape[0],)`` is used.
     dtype : :obj:`str`, optional
         Type of elements in input array. Defaults to ``None``.
     base_comm : :obj:`mpi4py.MPI.Comm`, optional
@@ -42,20 +48,38 @@ class MPILinearOperator:
 
     """
 
-    def __init__(self, Op: Optional[LinearOperator] = None, shape: Optional[ShapeLike] = None,
-                 dtype: Optional[DTypeLike] = None, base_comm: MPI.Comm = MPI.COMM_WORLD):
-        if Op:
+    def __init__(
+        self,
+        Op: Optional[Union[LinearOperator, MPILinearOperator]] = None,
+        shape: Optional[ShapeLike] = None,
+        dims: Optional[ShapeLike] = None,
+        dimsd: Optional[ShapeLike] = None,
+        dtype: Optional[DTypeLike] = None,
+        base_comm: MPI.Comm = MPI.COMM_WORLD
+    ):
+        if Op is not None:
             self.Op = Op
-            dtype = self.Op.dtype if dtype is None else dtype
-            shape = self.Op.shape if shape is None else shape
-        if shape:
-            self.shape = shape
-        if dtype:
-            self.dtype = dtype
+            dtype = Op.dtype if dtype is None else dtype
+            shape = Op.shape if shape is None else shape
+            # Optional arguments
+            dims = getattr(Op, "dims", None) if dims is None else dims
+            dimsd = getattr(Op, "dimsd", None) if dimsd is None else dimsd
+
+        # Infer shape from dims/dimsd if not provided
+        if shape is None:
+            if dims is None or dimsd is None:
+                raise ValueError(
+                    "Must provide either 'shape' or both 'dims' and 'dimsd'"
+                )
+            shape = (int(np.prod(dimsd)), int(np.prod(dims)))
+        self.shape = shape
+        self.dims = (shape[1],) if dims is None else dims
+        self.dimsd = (shape[0],) if dimsd is None else dimsd
+        self.dtype = dtype
         # For MPI
         self.base_comm = base_comm
-        self.size = self.base_comm.Get_size()
-        self.rank = self.base_comm.Get_rank()
+        self.size = base_comm.Get_size()
+        self.rank = base_comm.Get_rank()
 
     def matvec(self, x: DistributedArray) -> DistributedArray:
         """Matrix-vector multiplication.
--- a/pylops_mpi/LinearOperator.py
+++ b/pylops_mpi/LinearOperator.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import numpy as np
 from mpi4py import MPI
 from typing import Callable, Optional
@@ -254,9 +256,19 @@ class MPILinearOperator:
 
         """
         if isinstance(x, MPILinearOperator):
-            return _ProductLinearOperator(self, x)
+            Op = _ProductLinearOperator(self, x)
+            self._copy_attributes(
+                Op,
+                exclude=['dims']
+            )
+            Op.dims = x.dims
+            return Op
         elif np.isscalar(x):
-            return _ScaledLinearOperator(self, x)
+            Op = _ScaledLinearOperator(self, x)
+            self._copy_attributes(
+                Op
+            )
+            return Op
         else:
             if x is None or x.ndim == 1:
                 return self.matvec(x)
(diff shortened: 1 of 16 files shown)


rohanbabbar04:
Made some changes to the Operator in handling the `dims`, `dimsd`, and `shape`. Also, introduced `dims` and `dimsd` for operators wherever necessary, not many changes in the implementation were required :slightly_smiling_face: ...Do you think we need to introduce setters with some checks, we do not have a situation where this is required right now?

mrava87:
> Made some changes to the Operator in handling the `dims`, `dimsd`, and `shape`. Also, introduced `dims` and `dimsd` for operators wherever necessary, not many changes in the implementation were required 🙂 ...Do you think we need to introduce setters with some checks, we do not have a situation where this is required right now?

Thanks @rohanbabbar04! Ideally I would suggest we follow the PyLops approach as closely as possible. There we do all the hard logic into `LinearOperator` including getters/setters for shape,dims,dimsd, so one does not need to set all of them  (shape can be inferred from dims/dimsd or viceversa with the assumption that if one provides shape dims is shape[1] and dimsd is shape[0] - operator assumed to work with flat arrays only) and checks are made when one is set to ensure consistency. See https://github.com/PyLops/pylops/blob/dev/pylops/linearoperator.py

Not a must, but when easy/possible, it avoids having codebases depart from each other, and here it seems a no brainer

pylops_mpi/LinearOperator.py
```diff
@@ -26,36 +28,58 @@ class MPILinearOperator:
...
     :class:`pylops_mpi.DistributedArray` with ``Partition.SCATTER``, it is however
-    reccomended to use the :class:`pylops_mpi.basicoperators.MPIBlockDiag` operator
+    recommended to use the :class:`pylops_mpi.basicoperators.MPIBlockDiag` operator
     instead as this can also handle distributed arrays with subcommunicators.
 
     Parameters
     ----------
-    Op : :obj:`pylops.LinearOperator`, optional
-        PyLops Linear Operator to wrap. Defaults to ``None``.
+    Op : :obj:`pylops.LinearOperator` or :obj:`pylops_mpi.MPILinearOperator`, optional
```
mrava87: Do we ever pass a `MPILinearOperator` as into a `MPILinearOperator`?

pylops_mpi/LinearOperator.py
```diff
@@ -26,36 +28,58 @@ class MPILinearOperator:
...
-            dtype = self.Op.dtype if dtype is None else dtype
-            shape = self.Op.shape if shape is None else shape
-        if shape:
-            self.shape = shape
-        if dtype:
-            self.dtype = dtype
+            dtype = Op.dtype if dtype is None else dtype
+            shape = Op.shape if shape is None else shape
+            # Optional arguments
+            dims = getattr(Op, "dims", None) if dims is None else dims
```
mrava87: why None in getattr and not https://github.com/PyLops/pylops/blob/b6c103e2fd7a8862f9e2e4bafa1b273651c1aff0/pylops/linearoperator.py#L139?

pylops_mpi/basicoperators/BlockDiag.py
```diff
@@ -114,9 +114,10 @@ def __init__(self, ops: Sequence[LinearOperator],
         self.local_shapes_n = base_comm.allgather((self.nops, ))
         self.nnops = np.insert(np.cumsum(nops), 0, 0)
         self.mmops = np.insert(np.cumsum(mops), 0, 0)
-        shape = (base_comm.allreduce(self.nops), base_comm.allreduce(self.mops))
+        dimsd = (base_comm.allreduce(self.nops), )
```
mrava87, line 117: I guess this is the easy way... in pylops we also allow non-flat dims/dimsd if all ops have same dims/dimsd... and there the first element of dims/dimsd is the number of operators... i guess it may also work here because of the way we expect MPIBlockDiag to work, where each rank has a block and so an 'element' of that first axis?

Can postpose when a usecase needs it...

pylops_mpi/basicoperators/Halo.py
```diff
@@ -187,11 +187,11 @@ def __init__(
             int(np.prod(self.local_extent)),
         )
 
-        self.shape = (
+        shape = (
```
mrava87: mmh this should be the prime example where dims/dimsd makes sense, as this operator is fully built around the idea of turning 1D Distributed into ND and padding all all dimensions...

mrava87: Probably something like

```
dims = self.global_dims 
dimsd = self._allreduce(
            self.comm,
            None,
            self.local_extent,
        )
```
to be tested...

pylops_mpi/basicoperators/HStack.py
```diff
@@ -97,7 +97,7 @@ def __init__(self, ops: Sequence[LinearOperator],
             raise ValueError("Operators have different number of rows")
         hops = [oper.H for oper in self.ops]
         self.HStack = MPIVStack(ops=hops, base_comm=base_comm, dtype=dtype).H
-        super().__init__(shape=self.HStack.shape, dtype=self.HStack.dtype, base_comm=base_comm)
+        super().__init__(Op=self.HStack, dtype=self.HStack.dtype, base_comm=base_comm)
```
mrava87: mmh dont like/understand that... we pass pylops Op to MPILinearOperator so we dont need to reimplement matvec/rmatvec as it is always the same no matter the Op (it dispatches the local array to the matvec/rmatvec of Op)... but now you pass an MPILinearOperator and then still reimplement matvec/rmatvec, so what is the benefit... how does it differ to just pass self.HStack.dims/self.HStack.dims?

pylops_mpi/StackedLinearOperator.py
```diff
@@ -29,24 +29,40 @@ class of any new stacked operator developed within pylops-mpi.
...
     """
 
-    def __init__(self, shape: Optional[ShapeLike] = None,
-                 dtype: Optional[DTypeLike] = None,
-                 base_comm: MPI.Comm = MPI.COMM_WORLD):
-        if shape:
-            self.shape = shape
-        if dtype:
-            self.dtype = dtype
+    def __init__(
```
mrava87, line 45: Again ideally i would follow pylops approach, with setters/getters

mrava87 (changes requested):
@rohanbabbar04 great work!

Indeed in many cases it seems easy/natural to add support for dims/dimsd 😄 I just left some comments here and there, one major regarding the use of property and trying to avoid departing too much from PyLops when there is not an obvious reason to do so 😉

pylops_mpi/LinearOperator.py
rohanbabbar04: Yeah, just changed it should be `Op.shape[0/1]`

pylops_mpi/StackedLinearOperator.py
rohanbabbar04: Thanks, just add setters/getters

pylops_mpi/basicoperators/Halo.py
rohanbabbar04: Yeah, have introduced dims and dimsd instead of the shape

pylops_mpi/LinearOperator.py
rohanbabbar04: I think we don't need it as we can just pass the `dims` and `dimsd` like we do in `MPIHStack`.

pylops_mpi/basicoperators/BlockDiag.py
rohanbabbar04: Yeah, I think we can change it later one when we find an example/usecase.

pylops_mpi/basicoperators/HStack.py
rohanbabbar04: Yup, we can just pass the `dims` and `dimsd` directly. I have done that for this, so we don't need to pass Op.

pylops_mpi/basicoperators/Halo.py
```diff
@@ -186,12 +186,13 @@ def __init__(
...
-        shape = (
-            int(np.sum(self._local_extent_sizes)),
-            int(np.prod(self.global_dims)),
-        )
-        super().__init__(shape=shape, dtype=np.dtype(dtype), base_comm=comm)
+        dims = (self.global_dims,)
+        dimsd = (self._allreduce(
+            comm,
+            None,
+            np.prod(self.local_extent)
```
mrava87: mmh is `np.prod` correct? This reduces now to a scalar and then summed to other values in other ranks it remains a scalar, basically `shape[0]`... but if you have `self.local_extent` as I suggested the output should be still an array of the same size of dims with the n-dimensional shape of the padded array?

pylops_mpi/basicoperators/VStack.py
mrava87: What about this one? I guess same as for MPIVStack we stick so far with the flattened version?

mrava87 (changes requested):
@rohanbabbar04 great!

I have left a few minor comments, after that I think we should be good to go. 

Plus, one thing I just noticed, you did not add any test. I think we can again lean a bit on what we do in PyLops and maybe have distributed versions of these tests: `test_copy_dims_dimsd` in https://github.com/PyLops/pylops/blob/dev/pytests/test_linearoperator.py

pylops_mpi/basicoperators/VStack.py
rohanbabbar04: I agree!!, we keep the dims and dimsd (the flattened one)

pylops_mpi/basicoperators/Halo.py
rohanbabbar04: @mrava87 I think the shape would be different in both cases, an instance is an example below

```
# Running with 2 ranks
x = MPIHalo(dims=(32, 8), proc_grid_shape=(2, 1), halo=1)
```
the `self.local_extent` is (17, 8), if we apply `reduce-sum` it will be (34,16) which will be (544,), which is misleading as it should be (272,) (`reduce-sum of [(136,), (136, )]]`)

mrava87: @rohanbabbar04 Oh right, makes sense as whether something is divided into ranks or not depends on the `proc_grid_shape=(2, 1)`... here the second value is 1 so means axis 2 is not distributed, whilst axis 1 is distributed over the 2 ranks.

I would say that what you do is SAFE but not very useful as for dims/dimsd we want as much as possible to preserve the nd nature of the operator, but what I suggest in indeed WRONG.

We would need some sort of more fine-grained creation of dimsd where we reduce only the dimensions that are distributed and copy self.local_extent[I] for the i dimension(s) that are not (aka proc_grid_shape[i]=1.

For now we could keep what you have (like we did for MPIHStack/MPIVStack) but I just want to make sure you see this and we agree.

Please comment and resolve, after that we are ready to merge 😄

mrava87 (approved):
@rohanbabbar04 great, I think this is ready to go!

I just left a comment for you regarding `dimsd` in Halo but for now I think using the flattened version is ok 😄 

Feel free to merge 🚀

pylops_mpi/basicoperators/Halo.py
rohanbabbar04: Agreed, lets keep it to a flattened version for this PR. I will make certain changes when I am working on the N-D Distributed Arrays support for operators.

PR #202, opened 2026-06-23
Change GA rank  from 9 to 8

mrava87:


rohanbabbar04:
Some things which could potentially cause this as well:
- I think 9 would mean we are oversubscribing (since `num_proc` that is usually provided would be either 2, 4 or 8)
- MPICH uses a HYDRA scheduler that is different from how openmpi handles it, so could be a reason.
- I think we should change 9 to 8 as well, so that we follow the `2^n` norm, which is usually recommended when checking for oversubscribing.

mrava87:
@rohanbabbar04 thanks!

Yeah I was thinking about it myself (the choice of 9). Whilst it is true we want to make sure our codes allow any rank, choosing 9 is a really poor choice (i dont know any compute that has 9 cores - if you MPI also within the node - and why would anyone ever choose 9 anyways.

Let's try with 8 and see 😄

mrava87:
Still 82mins.... we should maybe try to figure out what makes this combination of parameters so slow (especially if you say is mostly due to ista/fista tests. 

For now, I approve and merge as 8 is anyways a better choice than 9

PR #203, opened 2026-07-12
Add local_array parameter to DistributedArray.

rohanbabbar04:
- Add local_array parameter to `DistributedArray`.
- Update ravel, conj and add reshape using the `local_array` parameter to create a view.

### DIFF
--- a/pylops_mpi/DistributedArray.py
+++ b/pylops_mpi/DistributedArray.py
@@ -131,6 +131,12 @@ class DistributedArray(DistributedMixIn):
         Broadcast, UnsafeBroadcast, or Scatter the array. Defaults to ``Partition.SCATTER``.
     axis : :obj:`int`, optional
         Axis along which distribution occurs. Defaults to ``0``.
+    local_array : :obj:`numpy.ndarray` or :obj:`cupy.ndarray`, optional
+        Existing local array to use as the underlying storage instead of
+        allocating a new one. Its shape must match the expected local shape
+        for the current rank. When provided, the local array's dtype and engine
+        (numpy or cupy) are used, and the ``dtype`` and ``engine`` arguments
+        are ignored.
     local_shapes : :obj:`list`, optional
         List of tuples or integers representing local shapes at each rank.
     mask : :obj:`list`, optional
@@ -146,6 +152,7 @@ class DistributedArray(DistributedMixIn):
                  base_comm: Optional[MPI.Comm] = MPI.COMM_WORLD,
                  base_comm_nccl: Optional[NcclCommunicatorType] = None,
                  partition: Partition = Partition.SCATTER, axis: int = 0,
+                 local_array: Optional[NDArray] = None,
                  local_shapes: Optional[List[Union[Tuple, Integral]]] = None,
                  mask: Optional[List[Integral]] = None,
                  engine: Optional[str] = "numpy",
@@ -158,10 +165,10 @@ class DistributedArray(DistributedMixIn):
         if partition not in Partition:
             raise ValueError(f"Should be either {Partition.BROADCAST}, "
                              f"{Partition.UNSAFE_BROADCAST} or {Partition.SCATTER}")
-        if base_comm_nccl and engine != "cupy":
+        self._engine = engine if local_array is None else get_module_name(get_array_module(local_array))
+        if base_comm_nccl and self._engine != "cupy":
             raise ValueError("NCCL Communicator only works with engine `cupy`")
-
-        self.dtype = dtype
+        self.dtype = dtype if local_array is None else local_array.dtype
         self._global_shape = _value_or_sized_to_tuple(global_shape)
         self._base_comm_nccl = base_comm_nccl
         if base_comm_nccl is None:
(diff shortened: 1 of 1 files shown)


rohanbabbar04:
With the introduction of `local_array`, we can now share memory instead of creating a copy.

pylops_mpi/DistributedArray.py
mrava87, line 162: Can you add a Raises section, just see we are having more and more, worth documenting them 😄

pylops_mpi/DistributedArray.py
```diff
@@ -810,16 +827,17 @@ def norm(self, ord: Optional[int] = None,
...
                                 axis=self.axis,
+                                local_array=local_array,
                                 local_shapes=self.local_shapes,
                                 mask=self.mask,
                                 engine=self.engine,
                                 dtype=self.dtype)
-        conj[:] = self.local_array.conj()
         return conj
 
     def copy(self):
```
mrava87, line 856: what about this one? maybe need to pass a copy of the local array otherwise changing one may change the other..

pylops_mpi/DistributedArray.py
```diff
@@ -852,17 +870,51 @@ def ravel(self, order: Optional[str] = "C"):
...
                                base_comm=self.base_comm,
                                base_comm_nccl=self.base_comm_nccl,
+                               local_array=local_array,
                                local_shapes=local_shapes,
                                mask=self.mask,
                                partition=self.partition,
                                engine=self.engine,
                                dtype=self.dtype)
-        local_array = np.ravel(self.local_array, order=order)
-        x = local_array.copy()
```
mrava87, line 864: this technically changes the behaviour because not only we avoid making a new array and then fill it (good!) but we also just pass the raveled `self.local_array` so technically the two underlying local arrays point to the same memory space... it may well be that this is what we want (really mimicking numpy `ravel`) and the fact no test fails tells me that we were maybe too conservative, but just checking you thought about this?

pylops_mpi/DistributedArray.py
```diff
@@ -852,17 +870,51 @@ def ravel(self, order: Optional[str] = "C"):
...
+        return arr
+
+    def reshape(self, local_shape: Tuple, axis: Optional[int] = 0):
+        """Return a reshaped DistributedArray
+
+        Parameters
+        ----------
+        local_shape : :obj:`tuple`
+            Shape of the local array on each MPI rank. For
+            :class:`Partition.SCATTER`, all local shapes must be identical on
```
mrava87: you say this, which is correct, but maybe worth also saying that for Partition.BROADCAST they instead need to all be the same... what happens if for some reason they are not, i guess there will be a numpy error at the local reshape step (i.e.,  self.local_array.reshape(local_shapes[self.rank]))?

pylops_mpi/DistributedArray.py
```diff
@@ -852,17 +870,51 @@ def ravel(self, order: Optional[str] = "C"):
...
+            Distribution axis in the reshaped global array. Defaults to ``0``.
+        """
+        local_shapes = self.base_comm.allgather(local_shape)
+        global_shape = list(local_shapes[0])
+        if self.partition is Partition.SCATTER:
+            ref_shape = local_shapes[0]
+            for i, shape in enumerate(local_shapes[1:], start=1):
+                if shape[:axis] != ref_shape[:axis] or shape[axis + 1:] != ref_shape[axis + 1:]:
+                    raise ValueError(
+                        f"All local shapes must match on every axis except axis={axis}. "
```
mrava87: maybe add `got {shape} in current rank and {ref_shape} in rank 0 instead...'

mrava87: Also would it not maybe better if just rank0 does this check and raises the error... eventually all ranks are checking the same thing and the error will be randomly coming by the one that is in front at that specific run instead of being consistently on the first rank that does not adhere to the rule?

mrava87 (changes requested):
@rohanbabbar04 great stuff!

I think this is a good addition and should help reduce the creation of additional numpy arrays when not strictly needed; just left a few minor comments for you to address before this is ready to go 😄

pylops_mpi/DistributedArray.py
rohanbabbar04: Yes, I have passed `self.local_array.copy()` for this.

pylops_mpi/DistributedArray.py
rohanbabbar04: I have added the Raises section.

pylops_mpi/DistributedArray.py
rohanbabbar04: Yup, I have added a check for Broadcast as well

pylops_mpi/DistributedArray.py
rohanbabbar04: Yes, I changed it so that each rank validates its own local shape against the local shape from rank=0. I wouldn't perform the check only on rank 0, as that could lead to deadlocks if rank 0 raises an exception while the remaining ranks continue execution and block on subsequent MPI operations.

pylops_mpi/DistributedArray.py
rohanbabbar04: Yes, this changes the behavior, with the behavior now closer to what we have in `numpy/cupy`(we let them decide), which, like you mentioned, is better. I don't think we require `ravel()` to create a new array, and potentially, from a user's perspective, they would want it to be a view as well.

**Similar to `ravel()`, the same applies to `reshape` and `conj`

mrava87 (approved):
@rohanbabbar04 this is good to go for me, feel free to merge

PR #207, opened 2026-08-11
Prepare release v0.7.0

rohanbabbar04:
- Update changelog for release `v0.7.0`
- Update pytest params to work only when mpi_rank>=2.


mrava87 (approved):
@rohanbabbar04 thanks for this, looks good to go for me 😀 

Just a curiosity: what was the rationale/need for the changes in the tests here?

rohanbabbar04:
No need, I just realised that for some tests we missed the mpi rank decorator, which we have for all the other test funcs. So to keep everything the same, better to add that to the tests which don't have them.

We only want to run the test when there is more than 1 rank.

PR #210, opened 2026-08-28
Bug: changes to suppress deprecation warning on float of 1-element array

mrava87:
# NumPy `ndim > 0` to-scalar deprecation: root cause and fix

## The warning

```
DeprecationWarning: Conversion of an array with ndim > 0 to a scalar is deprecated,
and will error in future. Ensure you extract a single element from your array before
performing this operation. (Deprecated NumPy 1.25.)
```

Triggered by any caller doing `float(...)` on the result of `DistributedArray.norm()`, because `norm()` never actually returned a scalar.

## Root cause

`DistributedArray.norm()` documented a `float` return but returned a **shape-`(1,)` array**:

1. `norm(axis=None)` flattens the local array and calls
   `_compute_vector_norm(..., axis=0, ord=ord)`.
2. That reduces via `_allreduce_subcomm`.
3. `mpi_allreduce` (`pylops_mpi/utils/_mpi.py:104`) always allocates
   `recv_buf = ncp.zeros(send_buf.size, dtype=...)` — a **1-D** buffer. For a scalar
   reduction `send_buf.size == 1`, so the result is `array([...])` with
   `ndim == 1, size == 1`.

`float()` on such an array is the deprecated implicit array-to-scalar conversion (an error in a future NumPy).

`DistributedArray.dot()` had the identical issue: it returned the raw size-1 allreduce buffer.

## The fix

New module-level helper in `pylops_mpi/DistributedArray.py`:

```python
def to_scalar(x) -> Union[bool, int, float, complex]:
    return x.item() if hasattr(x, "item") else x
```

- Backend-agnostic: `.item()` works for numpy and cupy arrays, and the
  `hasattr` guard covers the non-buffered-MPI paths that already return a
  plain scalar.
- **Dtype-preserving** — returns `float` in most cases, but for example
  a complex `vdot` stays `complex`.

Applied to:

| Location | Change |
|---|---|
| `DistributedArray.norm` (`axis=None` branch) | returns `float`; docstring notes scalar vs array per `axis` |
| `DistributedArray.dot` | returns `float`/`complex`; docstring return type fixed |
| `StackedDistributedArray.norm` | returns `float`; return annotation `bool \| float` -> `float` |

`norm(axis=...)` is unchanged and still returns an array.

`StackedDistributedArray.norm` **had** to change beyond the return value: it aggregated the per-array norms with `ncp.hstack(...)`, and since those are now host scalars, `cupy.hstack` on them would raise. It now aggregates with numpy (correct, because the inputs are host scalars).

## Call sites updated

`float` has no `.item()`, so every `.norm().item()` / `.dot(...).item()` chain had to be simplified.

- `pylops_mpi/optimization/cls_basic.py` — 10 spots across CG/CGLS. Note
  `a = float((self.kold / cOpc).item())` -> `float(self.kold / cOpc)` would have
  raised `AttributeError` otherwise (both operands are plain scalars now).
- `pylops_mpi/optimization/cls_sparsity.py` — 6 spots in ISTA/FISTA
  (`xupdate`, `costdata`, `costreg`).
- `pylops_mpi/optimization/eigs.py:86` — `b_k.dot(b1_k, vdot=True).item()` ->
  `b_k.dot(b1_k, vdot=True)`.
- `tests_nccl/test_distributedarray_nccl.py`, `tests_nccl/test_stackedarray_nccl.py`
  — the GPU tests called `.get()` on `dot()` / axis-less `norm()` results, which
  would now be `AttributeError` on a Python scalar. Removed on the 3 `dot` and 9
  `norm` assertions; **kept** on `norm(axis=...)`, which still returns a cupy array.

## Behaviour change to be aware of (cupy)

`norm(axis=None)` and `dot()` now return **host** scalars instead of device size-1 arrays, so they force a device-to-host sync. Nearly every call site already did `.item()` (same sync); the rest only divide a `DistributedArray` by the result, which is fine with a host scalar.

### DIFF
--- a/pylops_mpi/DistributedArray.py
+++ b/pylops_mpi/DistributedArray.py
@@ -4,19 +4,23 @@ from typing import Any, List, NewType, Optional, Self, Tuple, Union
 
 import numpy as np
 from mpi4py import MPI
-from pylops_mpi.Distributed import DistributedMixIn
 from pylops.utils import DTypeLike, NDArray
-from pylops.utils import deps as pylops_deps  # avoid namespace crashes with pylops_mpi.utils
+from pylops.utils import (
+    deps as pylops_deps,  # avoid namespace crashes with pylops_mpi.utils
+)
 from pylops.utils._internal import _value_or_sized_to_tuple
 from pylops.utils.backend import get_array_module, get_module, get_module_name
+
+from pylops_mpi.Distributed import DistributedMixIn
 from pylops_mpi.utils import deps
 
 cupy_message = pylops_deps.cupy_import("the DistributedArray module")
 nccl_message = deps.nccl_import("the DistributedArray module")
 
 if nccl_message is None and cupy_message is None:
-    from pylops_mpi.utils._nccl import nccl_asarray, nccl_split
     from cupy.cuda.nccl import NcclCommunicator
+
+    from pylops_mpi.utils._nccl import nccl_asarray, nccl_split
 else:
     NcclCommunicator = Any
 
@@ -39,8 +43,9 @@ class Partition(Enum):
     SCATTER = "Scatter"
 
 
-def local_split(global_shape: Tuple, base_comm: MPI.Comm,
-                partition: Partition, axis: int):
+def local_split(
+    global_shape: Tuple, base_comm: MPI.Comm, partition: Partition, axis: int
+):
     """To get the local shape from the global shape
 
     Parameters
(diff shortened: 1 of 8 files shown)


mrava87:
@rohanbabbar04 do you mind reviewing this 😄 

Note that:
- I was getting a lot of annoying warnings in the tests I am writing for the proximal operators/solvers, most of which are related to using `float(np.array)`, which is now a deprecated behavior... so I decided to see if I could fix them without making a lot of code changes
- the PR is mostly driven by Claude after we agreed on the approach (the new `to_scalar` method), so I am quite sure I did remove all `.item()` that we had scattered here and there as now norm/dot are by design always returning a pure float/complex scalar.
- the PR is quite noisy as we have recently introduced `.precommit` that operates on files with at least one change, and since our style was a bit all over the place, a lot of the changes are purely stylistic

rohanbabbar04 (approved):

PR #211, opened 2026-09-20
Add `py3.14t` and fix mpich issues in CI

rohanbabbar04:
- Add python-3.14t in the CI.
- Add UCX_TLS env var for mpich issues.


rohanbabbar04:
Have added `py3.14t`, better to run our tests on that as well (I think we can have that for `pylops` as well).

After some hours of debugging finally got the root cause of the Exit code 15(SIGTERM), I have added a new env var for Mpich which has now fixed it. Since the number of variables were increasing, I have shifted them to `env:`.

mrava87:
Thanks! Good addition 😄 

In PyLops we have already an action for free-threaded python: https://github.com/PyLops/pylops/blob/dev/.github/workflows/buildft.yaml. It is an action on its own as a few of our optional dependencies where not ready for ft yet once I set it up... should check again as it is a couple of months now since I did check last time.

mrava87 (approved):

rohanbabbar04:
> Thanks! Good addition 😄
> 
> In PyLops we have already an action for free-threaded python: https://github.com/PyLops/pylops/blob/dev/.github/workflows/buildft.yaml. It is an action on its own as a few of our optional dependencies where not ready for ft yet once I set it up... should check again as it is a couple of months now since I did check last time.

Oh great, didn't see that. I will merge this PR 🙂

